PR #20, opened 2019-06-12
Branch pipeline

vito:
This adds a branch pipeline and additional automation to set up the infrastructure and cloud config.

The pipeline does a full testing flow and includes jobs to deploy either the latest final Concourse version or the latest build from the specified branch.

In principle we could use this pipeline for ci.concourse-ci.org, configuring it with the 'master' branch. That's a decision for another day though!

### DIFF
--- a/deployments/branch-env/.gitignore
+++ b/deployments/branch-env/.gitignore
@@ -1,0 +1,2 @@
+keys/gcp.json
+.terraform/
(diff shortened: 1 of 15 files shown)


vito:
@YoussB Noted it would be nice to have a shorter path to getting a bosh deployed env from the specified branch, short-circuiting the testing flow. I think this makes sense, since sometimes we're just kicking the tires and don't expect a full ci/cd flow yet.

It would also be nice to avoid doing a BOSH blob bump + commit. Maybe we could just have a job that deploys with the latest built candidate, and another manual-only job that does the build + release creation + deploy all in one?

PR #23, opened 2019-06-17
Static fly build

kcmannem:
Zstd library needs to be used for fly execute. Zstd uses go bindings for the underlying C library. We need to turn CGO back on and pass it extra flags so it links everything statically.


kcmannem:
@pivotal-jamie-klassen thanks, didn't know you use just 1 dash.

PR #36, opened 2019-06-28
fix invalid YAML

pivotal-jamie-klassen:
and add a slack icon to the notify resource


pipelines/concourse.yml
```diff
@@ -1214,6 +1201,7 @@ resources:
 
 - name: notify
   type: slack-notifier
+  icon: slack
```
xtreme-james-thomson: nice

PR #49, opened 2019-07-11
break instead of exit

YoussB:
- fixed a gotcha that exited the whole suite


cirocosta:
Screen Shot 2019-07-11 at 12 23 52 PM

😥

PR #50, opened 2019-07-11
[k8s] remove "credhub" from helm distribution filter list

zoetian:
Concourse-ci filtered out the CREDHUB flags whereas helm now has these flags casuing the diff issue and the failing check-helm-param job in the pipeline


zoetian:
Opened an issue to implement the remaining `CREDHUB` flag: https://github.com/helm/charts/issues/15459

PR #51, opened 2019-07-11
Fix k8s-smoke 🤞

zoetian:
We missed the case when `fly` failed and `workers` result only contains a newline character (coming from linux/ unix command). This would pass the `-z` check, and "assume" there has worker started already. Hence, it would never `tick` and wait till tick counter reaches the `max-tick` and `exit 1`.

### DIFF
--- a/tasks/scripts/wait-worker
+++ b/tasks/scripts/wait-worker
@@ -47,8 +47,10 @@ try_until_responding() {
   echo "waiting for a worker to be running"
 
   while true; do
-    workers= fly -t $TARGET_NAME workers | grep 'running'
-    if [ -z workers ]; then
+    workers=$(fly -t $TARGET_NAME workers | grep 'running')
+    # if workers result is empty or only contains a newline
+    # print out dot and increase the tick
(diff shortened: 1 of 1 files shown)


YoussB:
nice catch, this might be the only error we should have handled 👍

tasks/scripts/wait-worker
```diff
@@ -47,8 +47,10 @@ try_until_responding() {
   echo "waiting for a worker to be running"
 
   while true; do
-    workers= fly -t $TARGET_NAME workers | grep 'running'
-    if [ -z workers ]; then
+    workers=$(fly -t $TARGET_NAME workers | grep 'running')
+    # if workers result is empty or only contains a newline
+    # print out dot and increase the tick
+    if [ -z "`echo $workers | tr -d '\n'`" ]; then
```
pivotal-jamie-klassen: when I'm in bash, I can do this:
```bash
bash-3.2$ [ -z $(echo -e '\n') ] && echo newlines dont count
newlines dont count
```

so I think it should be enough to just do

```bash
if [ -z $(fly -t $TARGET_NAME workers | grep 'running') ]; then
...
```

PR #54, opened 2019-07-19
Add charts-maintenance to reconfigure pipeline

### LINKED ISSUE #45: Move concourse-charts pipeline to prod and add failure notification
### Background:
- We have a [pipeline](https://hush-house.pivotal.io/teams/main/pipelines/concourse-charts) running under [hush-house](https://github.com/concourse/hush-house) that updates the merged branch in `concourse/charts` repo to be a merging of the `helm/charts` master branch and branches that we specify in [this file](https://github.com/concourse/charts/blob/maintenance/refs.json) so then we can use this branch as an RC for the Concourse chart when we have unmerged PRs.
- This pipeline has been failing silently for some time now resulting in `k8s-topgun` failures.

### Request:
- This pipeline should be moved to our prod environment, so it can be easily monitored.
- Also, would be handy to add a slack notification for that whenever it fails (or maybe adding it to the concourse pipeline itself 🤔)

pivotal-jamie-klassen:
fixes #45

### DIFF
--- a/pipelines/reconfigure.yml
+++ b/pipelines/reconfigure.yml
@@ -13,6 +13,20 @@ resources:
     paths:
     - pipelines
 
+- name: charts-maintenance
+  type: git
+  icon: github-circle
+  source:
+    uri: https://github.com/concourse/charts
+    branch: maintenance
+
+- name: maintenance-image
+  type: registry-image
+  source:
+    password: ((docker.password))
+    repository: concourse/charts-maintenance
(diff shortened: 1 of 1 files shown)


cirocosta:
ooooh nice! thank you!

cirocosta:
sorry for not having merged it in at the time of approval 😬  it's not not up to date - would you mind updating it?

Thank you!

PR #55, opened 2019-07-22
update bpm-release url

vito:
this should fix the invalid shas in concourse-bosh-deployment


pivotal-jamie-klassen:
@vito looks like this needs a signoff

vito:
lazily used my admin passwords to skip it

PR #56, opened 2019-07-24
Add `restore` step to backup pipeline

cirocosta:
As very well pointed out by @pivotal-jamie-klassen, backups are only as
good as their restore.

This commit represents an effort to have that second half (the restore
part) as the subsequent job right after the backup.

In overall, the pipeline for backup + restore looks like this:

	BACKUP

		1. take a backup from prod using BBR
		2. store the backup in GCS

	RESTORE

		1. generate ops files that match the versions in prod
		2. deploy concourse having just a database
			2.1 with `superuser` to concourse so we can enable
			    `pg_statements`
		3. `bbr restore` against it
		4. pause all pipelines
		5. bring `web` and `worker` up
		6. trigger a job in a pipeline that already existed before

This way, we're able to verify not only that `bbr` is able to put the
bits back into a database, but that concourse is able to use it
accordingly too.

- fixes https://github.com/concourse/concourse/issues/2094

---

What's missing:

- [x] validate the `trigger-job` step
- [x] bring back the `ensure` that deletes the deployment afterwards
- [x] have the `passed` constraints in `restore`
- [x] add the `midnight` resource back

These are small things, but the restore takes quite a while 😅  so the iterations are very slow.

Thank you!


cirocosta:
Hey, got a successful run after finishing those final steps that I had in the checklist - it's looking like this:

Screen Shot 2019-07-25 at 8 13 13 AM

I think the `restore` job looks quite self-explanatory now in terms of nomenclature:

Screen Shot 2019-07-25 at 8 13 26 AM
 
thanks!

tasks/bbr-trigger-existing-pipeline.yml
```diff
@@ -0,0 +1,36 @@
...
+      echo "waiting for /api/v1/info..."
+      sleep 2
+    done
+
+    echo "installing fly..."
+    curl "$web_url/api/v1/cli?platform=linux&arch=amd64" -o /usr/local/bin/fly
+    chmod +x /usr/local/bin/fly
+
+    fly -t target login -c "$web_url" -u admin -p admin
+    fly -t target unpause-pipeline -p bbr-test-2
```
taylorsilva: Can we give the `bbr-test-2` pipeline a different name, one that doesn't make it seem temporary. Worried someone might delete it. We could also add it to the `reconfigure-pipeline` to help ensure it's persistence.

taylorsilva: Can we add the pipeline yaml for `bbr-test-2` in this repo too?

tasks/discover-bosh-endpoint-info.yml
```diff
@@ -7,7 +7,6 @@ image_resource:
     repository: concourse/unit
 
 inputs:
-- name: concourse
```
taylorsilva: Can we put this in another PR, seems unrelated.

PR #57, opened 2019-07-25
Add k8s-topgun-pks job

taylorsilva:
This adds a job running topgun/k8s test running on PKS in vsphere.

We created an external worker that's connected to prod (ci.concourse-ci.org). This external worker is located in the vsphere bosh environment where our PKS cluster is located. The vsphere environment is locked down to pivotal network access only which is why we needed this worker. We can't publicly access the PKS cluster or bosh director. 

CC @zoetian

### DIFF
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -34,6 +34,7 @@ groups:
   - k8s-check-helm-params
   - k8s-smoke
   - k8s-topgun
+  - k8s-topgun-pks
 
 - name: bosh
   jobs:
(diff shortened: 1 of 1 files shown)


pipelines/concourse.yml
```diff
@@ -426,6 +427,44 @@ jobs:
...
+      params: {format: oci}
+    - get: unit-image
+      passed: [k8s-smoke]
+      trigger: true
+      tags: [pks]
+    - get: charts
+      trigger: true
+      passed: [k8s-smoke]
+      tags: [pks]
+    - get: ci
```
YoussB: Why aren't we using the `tags` here as well?

YoussB:
Looks Good, just added a small comment 👍

pipelines/concourse.yml
taylorsilva: Missed it there, added it now

PR #59, opened 2019-07-25
[#4141] tagging netgo to be used while compiling with cgo

kcmannem:


kcmannem:
we confirmed that this worked on our local machine and asked a commenter on the issue to compile them selves to see if they were still facing the issue. It seems to be resolved by providing this `--tags netgo` flag.

https://github.com/concourse/concourse/issues/4141#issuecomment-514541431

tasks/scripts/fly-build
```diff
@@ -21,10 +21,10 @@ pushd concourse
   export CGO_ENABLED=1
 
   if [ "$platform" = "linux" ]; then
-    ldflags+=' -linkmode external -extldflags "-static"'
+    ldflags+=' -extldflags "-static"'
   fi
 
-  go build -ldflags "$ldflags" -o $output/fly ./fly
+  go build --tags netgo -ldflags "$ldflags" -o $output/fly ./fly
```
cirocosta: Hey,

Should we force rebuilding of all packages too? i.e., leveraging `-a` in `go build`?

```console
$ go help build
...
The build flags are shared by the build, clean, get, install, list, run,
and test commands:

	-a
		force rebuilding of packages that are already up-to-date.
...
```

I always use in my build scripts, but I'm honestly not sure if it's really needed or not 🤷‍♂ 

wdyt?
thx!

kcmannem: wouldn't hurt but because we build in a container wouldn't everything need to be rebuilt anyway?

cirocosta: hmmmm it wouldn't build the "core" ones

```bash
mkdir /tmp/sample && cd /tmp/sample
echo "package main
func main () {
  i := 0
  i += 1
}" > ./main.go
```

then, perform the regular build

```bash
go build -v .
_/tmp/sample
```

then, another with `-a`:

```bash
go build -v -a .
internal/cpu
runtime/internal/sys
runtime/internal/atomic
runtime/internal/math
internal/bytealg
runtime
_/tmp/sample
```

Does that matter or not? I don't know 🤷‍♂

kcmannem: makes sense to me, I'll add it in

kcmannem:
the cf cli use these flags `REQUIRED_FOR_STATIC_BINARY =-a -tags netgo -installsuffix netgo` , im not really sure what -installsuffix does but I'll update the pr with/without it after I find out.

tasks/scripts/fly-build
kcmannem: done

PR #65, opened 2019-07-31
wait for cloud-init before apt-get update

pivotal-jamie-klassen:
cloud-init will set up the correct apt sources, so we can avoid broken packages.


vito:
ahhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh is this what causes the postgres install flakes?

PR #71, opened 2019-07-31
Make yaml yml

cirocosta:
After the addition of `make` and change to the extension used in that pipeline filename (see https://github.com/concourse/charts/commit/781e0b8f305d54db13cce3eb12e2aa764de115e0), we can reference the right name 👍 

Fixes #54


pivotal-jamie-klassen:
correction: 'make yaml yml again'

PR #72, opened 2019-07-31
Docker tag names can't contain /

xtremerui:


pipelines/reconfigure.yml
```diff
@@ -115,7 +115,7 @@ jobs:
         team: main
         config_file: pipelines/pipelines/branch.yml
         vars:
-          branch_name: feature/3788-checks
+          branch_name: checks
```
jomsie: branch name getting cooler...

PR #86, opened 2019-08-06
remove jobs for testing

xtremerui:
for fast deploying drills env.

### DIFF
--- a/pipelines/drills.yml
+++ b/pipelines/drills.yml
@@ -1,0 +1,895 @@
+# the following vars must be specified on configuration:
+#
+#   ((branch_name)): the branch of 'concourse' and 'concourse-bosh-release'
+#   ((environment)): a name for the bosh-deployed drills testing environment
+#   ((admin_username)),
+#   ((admin_password)): credentials for the admin user
+#
+# the branch name must not have any funny characters, and must exist on the
+# following repos:
+#
+#   concourse/concourse
+#   concourse/concourse-bosh-release
+#
+# the drills environment will be reachable at the following url:
+#
+#   https://((environment)).concourse-ci.org
+
+resource_types:
+- name: gcs
+  type: registry-image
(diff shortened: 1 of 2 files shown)


xtremerui:
hmm wait for this one I think the discussion is still going.

vito:
:thinking: I don't like this from the standpoint that I expected the branch pipeline to be used for long-lived branches, i.e. features that are in development for the next major release, where we have made a conscious choice to not integrate with `master` yet so we can keep shipping minor releases until it's ready. The new algorithm + build re-triggering are a prime example, as they're slated for v6.0.

It's true that `testflight`/`watsjs` already run for PRs, but `upgrade`/`downgrade` don't, and have caught legitimate breakage already. I also think it's nice to have one place to look for build status. Sometimes we don't even have a PR open for the branch until it's 'ready to go'. We can change that and submit PRs early and often I guess.

So, my main concern I guess is that we're changing the entire use case for this template, and I don't think the need for the original use case has gone away. It feels like the tension is coming from the fact that the template is being used for both long-lived branches and as a way to get a quick feedback loop going with a dev environment. Maybe those concerns should be separated into separate templates instead?

Another alternative is to just eliminate the `passed` constraints. This would give you your fast-path-to-dev-environment while asynchronously providing feedback via the other jobs. That would probably be sufficient for long-lived branch pipelines too.

vito:
We could also maybe run `upgrade`/`downgrade` for PRs. :thinking:

ddadlani:
@vito that makes sense. I do think that the `branch` pipeline would be easier to use without the extra checks but eliminating the passed constraint is sufficient

pivotal-jwinters:
We've created a new `drills.yml` that people can use if they don't want to run `testflight`, `watsjs`, `upgrade`, `downgrade`.

vito:
sweet! thanks y'all

PR #92, opened 2019-08-08
Have Wings in its own pipeline

cirocosta:
Hey,

Here we're separating the deployment of Wings into its own pipeline so that we can:

- easily pin versions without affecting the whole development pipeline
- leverage any release candidates regardless of which pipeline generated them

Thank you!

### DIFF
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -46,7 +46,6 @@ groups:
   - bosh-smoke
   - bosh-topgun
   - bosh-prod-deploy
-  - bosh-wings-deploy
   - bosh-check-props
   - merge-cbd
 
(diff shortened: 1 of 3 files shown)


jomsie:
awesome!!

should this new pipeline also be in the reconfigure pipeline https://github.com/concourse/ci/blob/master/pipelines/reconfigure.yml

and i'm assuming HH will go the same way as this, but that's a separate issue!

thanks!

cirocosta:
good point! we just updated it to have it there 👍

jomsie:
also @cirocosta @xtremerui, i just restarted `reconfigure-pipelines` again so it reconfigured the new wings pipeline successfully

PR #98, opened 2019-08-13
refactor `concourse` pipeline to build and ship separate ubuntu and alpine images

cirocosta:
Hey,

With the goal of having a `concourse/concourse` image that is based of Ubuntu-based resources types, I end up refactoring the pipeline in a way that would allow us to consume them without getting into hacks like we did for the BOSH release where we'd swap contents from the tarball with resource types that we'd get from GitHub.

The pipeline is currently "set" at https://ci.concourse-ci.org/teams/main/pipelines/concourse-beta2?group=develop .

Here's a breakdown of the final result:

1. `dev-image` was broken into `resource-types` and `dev-image`, with the former consuming a single flavor of `resource-types` (alpine)

dev-image-using-alpine

you can see how we generate multiple flavors:

resource-types-two-images

2. made `build-rc` produce two flavors of `rc`s. 

This allows us to not have to have the resources floating around the pipeline, as well as be able to fetch the RC corresponding to whatever flavor you want

rc-interaction

3. made `shipit` ship both images: `alpine` and `ubuntu

publishing-two-images

I didn't run the pipeline yet - just the compact one, which was working as intended.

Wdyt?

Any thoughts?

Thank you!

---

ps.: there was an old comment here detailing how I was thinking about tackling the issue. If you'd like to see it, click below:

<details>
<summary>
<e>click on me!11!!!</e>
</summary>

The approach that we went for is:

1. separate `dev-image` from `resource-types`
2. consume in `dev-image` the `alpine`-based resource types
3. produce `alpine` **and** `ubuntu` linux-rc's
  - this way we keep track of the exact tarball with the resource types that got to the BOSH release
4. produce `alpine` **and** `ubuntu` `concourse/concourse` images
  - this way those not wanting to leverage `ubuntu` don't get hurt by the size of `ubuntu`-based resource types
5. generate a list of the debian packages that are present in the final image
  - so we can scan for non-compliant packages and CVEs

Oh, and this will get rid that ugly bunch of resource types getting dumped into `bosh-bump` (where we do the resource types swapping).

I'm marking the PR as a draft as the development was done in a separate pipeline (`concourse-compact`) that just involves the places where we interact with either resource types or images (for faster iteration cycles).

The compact one looks like this right now:

Screen Shot 2019-08-13 at 12 14 52 PM

I'll work next on integrating what we learned with it into the main pipeline 👍 

Please let me know if you have any thoughts/opinions about it!

Thanks! 
</details>


cirocosta:
(btw, the idea is to squash all of those `wip`s)

cirocosta:
(if you're wondering how big is the difference between those two images: 443MB)

cirocosta:
Here's a breakdown of the final result:

1. `dev-image` was broken into `resource-types` and `dev-image`, with the former consuming a single flavor of `resource-types` (alpine)

dev-image-using-alpine

you can see how we generate multiple flavors:

resource-types-two-images

2. made `build-rc` produce two flavors of `rc`s. 

This allows us to not have to have the resources floating around the pipeline, as well as be able to fetch the RC corresponding to whatever flavor you want

rc-interaction

3. made `shipit` ship both images: `alpine` and `ubuntu

publishing-two-images

I didn't run the pipeline yet - just the compact one, which was working as intended.

Wdyt?

Any thoughts?

Thank you!

pipelines/concourse.yml
```diff
@@ -1679,3 +1759,40 @@ resources:
...
+    tag: ubuntu
+    username: ((docker.username))
+    password: ((docker.password))
+
+- name: concourse-rc-image-alpine
+  type: registry-image
+  icon: *image-icon
+  source:
+    repository: concourse/concourse-rc
+    tag: switch-base
```
cirocosta: oops, this should actually be `latest` or just nothing (as `alpine`-based concourse images would still be the default)

pipelines/concourse.yml
```diff
@@ -1679,3 +1759,40 @@ resources:
...
+    username: ((docker.username))
+    password: ((docker.password))
+
+- name: concourse-rc-image-ubuntu
+  type: registry-image
+  icon: *image-icon
+  source:
+    repository: concourse/concourse-ubuntu-rc
+    tag: switch-base
```
cirocosta: oops should be `ubuntu`

pipelines/concourse.yml
```diff
@@ -1679,3 +1759,40 @@ resources:
...
+    repository: concourse/concourse-rc
+    tag: switch-base
+    username: ((docker.username))
+    password: ((docker.password))
+
+- name: concourse-rc-image-ubuntu
+  type: registry-image
+  icon: *image-icon
+  source:
+    repository: concourse/concourse-ubuntu-rc
```
cirocosta: should be just `concourse/concourse-rc`

PR #104, opened 2019-08-15
parametrize the linker type for fly builds

### LINKED ISSUE #103: add the option to run linux cgo builds using an external linker
for example, the go linker can't handle pthreads (a transitive
dependency of v5.2.x), whereas the external linker seems to be
able to statically compile it.

pivotal-jamie-klassen:
fixes #103, hopefully unblocking the 5.2.1 release.


pivotal-jamie-klassen:
@kcmannem I tested it [here](https://ci.concourse-ci.org/builds/6731089)

PR #106, opened 2019-08-15
release: pin charts version to match Charts' concourse versions

cirocosta:
As the Chart contains the fields that are specified for any given
version (e.g., `master` currently points to 5.4.1), when leveraging
`branch: master` in the release pipeline for 5.2., we end up with
problems given that the fields differ.

Thus, this commit pins the version of `helm/charts` to the last commit
where the version of Concourse appropriately matches.

Thanks!

### DIFF
--- a/pipelines/reconfigure.yml
+++ b/pipelines/reconfigure.yml
@@ -75,6 +75,7 @@ jobs:
         vars:
           release_major: "5"
           release_minor: "5.2"
+          charts_version: '060b57708ed34ffd47f8076bf5906167dfce96b8'
           concourse_smoke_deployment_name: "concourse-smoke-5-2"
           use_external_linker: true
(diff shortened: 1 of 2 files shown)


pivotal-jamie-klassen:
@cirocosta I think this will unblock the 5.2.1 release, but I'm wondering if it's worth parametrizing something like this? since future release pipelines might also need to pin the version of the helm chart they are working with?

cirocosta:
yeahh .. that's a good point

I've been wondering about the possibility of actually getting out of `helm/charts` completely and rolling our chart from `concourse/chart` 🤔  

That'd allow us to have release branches like in the other repos, allowing us to fix problems with older versions that we still support (something that is not possible to do right now with the `helm/charts` repo as it's "move-forward-only" 😬 )

PR #107, opened 2019-08-15
use resource-types image in release pipeline

pivotal-jamie-klassen:


tasks/scripts/docker-semver-tags
```diff
@@ -3,26 +3,40 @@
 set -e -u
 
 version=$(cat version/version)
+if [ ! -z "$SUFFIX" ]; then
```
pivotal-jamie-klassen: oops, this has to move down...

xtremerui: right, per my comment, how could a reviewer help in this case

pivotal-jamie-klassen:
this is the deepest question of all. in some sense we should have some automated way of setting the pipeline up in a staging environment, automatically converting all the sensitive resources into reasonable fakes. i usually just set the pipeline locally to see if it will validate, and otherwise it just goes straight into production and we figure it out as we go.

PR #111, opened 2019-08-15
use resource-types-alpine-image instead of building it in drills

xtremerui:


cirocosta:
oh wait, I think there's something wrong

xtremerui:
Screen Shot 2019-08-17 at 12 40 50 AM

@cirocosta how about now?

cirocosta:
ooh coool! sounds good to me now!

PR #112, opened 2019-08-19
pipeline: add a pipeline to clean up prod bosh environment.

pivotal-bin-ju:
This commit is for the issue: https://github.com/pivotal/concourse-ops/issues/115

We added the pipeline `bosh-prod-cleanup` to run `bosh clean-up --all` every Saturday.


cirocosta:
Hey @pivotal-jamie-klassen , I just added `bosh-cleanup` to the reconfigure pipeline. Would you mind taking another look at it?

Thanks!

jomsie:
Just for my own reference since I've not used bosh clean-up before,

```
Clean-Up¶
bosh -e my-env clean-up [--all]

Cleans up unused resources but keeps orphaned disks and the two most recent versions of stemcells and releases.

--all clean up all unused resources including orphaned disks.
```

Seems like a good thing to run every-so-often!

PR #113, opened 2019-08-20
minor pipeline improvements

pivotal-jamie-klassen:
* track image used to run unit tests on PRs via resource
* no more ((vars)) directly in task configs

### DIFF
--- a/pipelines/prs.yml
+++ b/pipelines/prs.yml
@@ -60,6 +60,11 @@ resources:
   icon: docker
   source: {repository: concourse/dev}
 
+- name: unit-image
+  type: registry-image
+  icon: docker
+  source: {repository: concourse/unit}
+
 - name: postgres-image
   type: registry-image
   icon: docker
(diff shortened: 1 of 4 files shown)


pipelines/prs.yml
```diff
@@ -60,6 +60,11 @@ resources:
   icon: docker
   source: {repository: concourse/dev}
 
+- name: unit-image
```
cirocosta: I was wondering the other day about renaming this image 🤔  At least for me, the name `unit` doesn't convey much what it really brings (a common set of tools that many tasks need).

pivotal-jamie-klassen: yeah, I'd be happy to revisit the naming convention around `unit` and `dev`, but I have no bright ideas at this moment.

pivotal-jamie-klassen: yeah, I'd be happy to revisit the naming convention around `unit` and `dev`, but I have no bright ideas at this moment.

kcmannem: sameer and I were confused as well. We were thinking somthing like `test-tools` or `test-utils`

pivotal-jamie-klassen: #118

PR #114, opened 2019-08-20
concourse/#4302 updating dev/unit base image to golang-builder

kcmannem:
Update `concourse/unit` and `concourse/dev` to use the `concourse/golang-builder` as they previously were based on the official golang image which had a major bump from `stretch` to `buster`. 

The main concourse pipeline experienced high CPU usage as well as timeouts for tests that utilized the `buster` variant.

Fixes concourse/concourse#4302
Co-authored-by: Krishna Mannem <kmannem@pivotal.io>

### DIFF
--- a/dockerfiles/dev/Dockerfile
+++ b/dockerfiles/dev/Dockerfile
@@ -6,7 +6,7 @@
 # * warming the module cache
 # * warming the build cache
 
-FROM golang:1
+FROM concourse/golang-builder
 
 RUN apt-get update && apt-get -y install \
       iproute2 \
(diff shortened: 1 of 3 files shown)


cirocosta:
Hey,

Could we also tie the results of building `concourse/golang-builder` to `dev` to kick off the pipeline from there whenever we update `golang-builder`?

E.g., we could take as an `ARG` the digest of `golang-builder`, which could be fed into `concourse/dev` 🤔  

```Dockerfile
ARG ref=":latest"

FROM concourse/golang-builder${ref}
```

so that by default, it uses `latest`, but when running in the pipeline, it could take the digest from the `version` file that `regstry-image-resource` brings, which would be similar to

```sh
docker build -t myimage --build-arg ref='@sha256:blabla'
```

Wdyt?

Thanks!

xtreme-sameer-vohra:
Good point @cirocosta 

Updated the PR to add triggers to dev & unit images.

PR #120, opened 2019-08-22
Get release notes from concourse/concourse

jomsie:
- concourse and release pipelines now get the release notes from concourse/release-notes
- moved other pipeline (prep-release-assets) from release notes into ci

Please review concourse/concourse#4312 first!

### DIFF
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -1117,14 +1117,14 @@ jobs:
       passed: [shipit]
     - get: darwin-rc
       passed: [shipit]
-    - get: release-notes
     - get: docs
+    - get: ci
   - in_parallel:
     - task: prep-release-assets
-      file: release-notes/ci/prep-release-assets.yml
+      file: ci/tasks/prep-release-assets.yml
       image: unit-image
     - task: build-release-notes
-      file: release-notes/ci/build.yml
+      file: ci/tasks/build-release-notes.yml
       image: unit-image
       params:
         GITHUB_TOKEN: ((concourse_github_dummy.access_token))
@@ -1571,13 +1571,6 @@ resources:
   icon: *git-icon
   source:
     uri: https://github.com/concourse/docs
-    branch: master
-
-- name: release-notes
-  type: git
-  icon: *git-icon
-  source:
-    uri: https://github.com/concourse/release-notes
     branch: master
 
 - name: docs-gh-pages
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -860,14 +860,14 @@ jobs:
       passed: [shipit]
     - get: darwin-rc
       passed: [shipit]
-    - get: release-notes
     - get: docs
+    - get: ci
   - in_parallel:
     - task: prep-release-assets
-      file: release-notes/ci/prep-release-assets.yml
+      file: ci/tasks/prep-release-assets.yml
       image: unit-image
     - task: build-release-notes
-      file: release-notes/ci/build.yml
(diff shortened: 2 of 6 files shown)


tasks/build-release-notes.yml
```diff
@@ -0,0 +1,17 @@
...
+platform: linux
+
+image_resource:
+  type: registry-image
+  source: {repository: concourse/unit}
+
+inputs:
+- name: concourse
+- name: version
+  optional: true
```
pivotal-jamie-klassen: you need another input `name: ci`

tasks/build-release-notes.yml
```diff
@@ -0,0 +1,17 @@
...
+
+outputs:
+- name: built-notes
+
+run:
+  path: scripts/build-release-notes
```
pivotal-jamie-klassen: this should be `path: ci/tasks/scripts/build-release-notes`

tasks/scripts/build-release-notes
```diff
@@ -0,0 +1,16 @@
...
+elif [ "$#" -eq 1 ]; then
+  version=$1
+else
+  echo "version must be specified, either as an argument or an input" >&2
+  exit 1
+fi
+
+echo "$version" > ./built-notes/release-name
+# if this fails, it's because we forgot to manually promote latest.md to $version.md
+cp concourse/release-notes/"$version".md ./built-notes/notes.md
```
pivotal-jamie-klassen: i think we can (and should) do this with an automated commit -- the constraint on whether a version is releasable should be controlled by `passed:` constraints in the pipeline, and something feels awkward to me about making a 'release commit' up front without first knowing whether the code in question is gonna pass... I can imagine some kind of deadlock where we think we can release, we promote the `latest.md` and then it turns out our code is broken, but by the time the next RC passes testing we've merged more features so our release notes have to change! and the cycle goes on.

However, the way it is here isn't gonna directly break anything, so let's save ^this work for another PR.

pipelines/release.yml
```diff
@@ -1138,13 +1138,6 @@ resources:
     uri: https://github.com/concourse/docs
     branch: master
 
-- name: release-notes
-  type: git
-  icon: *git-icon
-  source:
-    uri: https://github.com/concourse/release-notes
-    branch: master
-
```
pivotal-jamie-klassen: :+1: very pleasing.

pipelines/concourse.yml
```diff
@@ -1573,13 +1573,6 @@ resources:
     uri: https://github.com/concourse/docs
     branch: master
 
-- name: release-notes
-  type: git
-  icon: *git-icon
-  source:
-    uri: https://github.com/concourse/release-notes
-    branch: master
-
```
pivotal-jamie-klassen: 👍 very pleasing.

tasks/build-release-notes.yml
deniseyu: d'oh. thanks for the catch

tasks/scripts/build-release-notes
deniseyu: yea, we were trying to come up with a process that encouraged a human to look over the contributor-submitted release notes as a whole and check that they're readable, but this workflow is clunky for sure. also tricky that we can't have the pipeline push directly to master and do this rename + push as part of the manually triggered release job from within the pipeline!

pivotal-jamie-klassen: maybe the testing jobs could first check to see if the latest commit is only a documentation change, and automatically pass in that case...

PR #122, opened 2019-08-23
concourse/release: add dpkg list generation

cirocosta:
As a way of having the results of `dpkg -l` stored for the ubuntu-base
images that we ship, this commit adds two extra steps to the
publishing/releasing:

1. dpkg list generation
2. uploading of the result of `dpkg -l` to our bucket

Naturally, as `dpkg -l` runs in a container leveraging the container
image that was previously built, we need to perform an extra `get` that
retrieves the image in the `rootfs` (concourse-specific) format, rather
than `oci` (used to push to a registry).


cirocosta:
(sorry for the "force pushes" 😅)

tasks/generate-dpkg-list.yml
```diff
@@ -0,0 +1,28 @@
...
+
+inputs:
+- name: version
+  optional: true
+
+outputs:
+- name: dpkg-list
+
+params:
+  COMPONENT_NAME: ~
```
jomsie: just a question (for my own knowledge!) why the `~`'s?

PR #134, opened 2019-08-30
make dumb-init executable

pivotal-jamie-klassen:


jomsie:
oops forgot to say

NICE SNAG!!

PR #137, opened 2019-08-31
Fix release notes flow

pivotal-jamie-klassen:
Now we shouldn't need to wait for a docs-only change to pass through the whole
pipeline just to ship!

NOTE: this means that master will always be the source of truth for release
notes, even if the release itself comes from a different branch.


pipelines/release.yml
```diff
@@ -991,14 +992,25 @@ resources:
   source:
     uri: https://github.com/concourse/concourse.git
     branch: release/((release_minor)).x
+    ignore_paths:
+    - release-notes/
 
 - name: ci
   type: git
-  icon: &git-icon github-circle
+  icon: *git-icon
```
cirocosta: whops good catch

pipelines/release.yml
```diff
@@ -991,14 +992,25 @@ resources:
...
   source:
     uri: https://github.com/concourse/ci.git
     branch: master
 
+- name: release-notes
+  type: git
+  icon: *git-icon
+  source:
+    uri: https://github.com/concourse/concourse.git
+    branch: master
```
cirocosta: is this `master` because we'd write release notes for, say, `5.2.4` on `master`, rather than on the `5.2.x` branch?

pivotal-jamie-klassen: Exactly. If there's any single place where you'd expect to find all the release notes, that folder on master seems like the natural place. Id rather that than have the release notes scattered among multiple branches, but that's a subjective preference.

PR #140, opened 2019-09-04
remove label so semver prs will be tested

pivotal-jamie-klassen:
having an approved-for-ci label on the semver resource prevented pr tests to run


jomsie:
Lgtm

PR #142, opened 2019-09-05
fix norsk status reporting

### LINKED ISSUE #145: slack notification on OSL failure

pivotal-jamie-klassen:
task was missing input
also, fixes #145

### DIFF
--- a/pipelines/norsk-mirror.yml
+++ b/pipelines/norsk-mirror.yml
@@ -1,4 +1,9 @@
 ---
+resource_types:
+- name: slack-notifier
+  type: registry-image
+  source: {repository: mockersf/concourse-slack-notifier}
+
 resources:
 - name: latest-scan-status
   type: git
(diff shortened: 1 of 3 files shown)


cirocosta:
Good catch!

I noticed that the pipeline is not currently in ` ci/pipelines/reconfigure.yml` though - I guess that can be done in a different PR if desired 😁  

thanks!

pipelines/norsk-mirror.yml
```diff
@@ -28,6 +41,32 @@ jobs:
   - task: check-status-file
     file: ci/tasks/check-norsk-status.yml
     input_mapping: {scan-status: latest-scan-status}
+  on_success: &fixed-status
```
cirocosta: if we're not using these anchors, I think it might be better removing them 🤔  e.g., it's easy to think that the this object is the same as `check-p-concourse-4.2.x`'s on-success given that the anchors indicate as if they were the same (although we're not dereferencing this one over there)

wdyt?

thx!

pivotal-jamie-klassen: haha yes good catch!

pivotal-jamie-klassen:
> Good catch!
> 
> I noticed that the pipeline is not currently in ` ci/pipelines/reconfigure.yml` though - I guess that can be done in a different PR if desired 😁
> 
> thanks!

Added it!

PR #152, opened 2019-09-06
Add credhub and credhub-prefix ops-files

taylorsilva:
Diff looks a bit messy because I decided to sort the ops files, but all this PR does is add two ops-files to enable credhub for wings.

- https://github.com/concourse/concourse-bosh-deployment/blob/master/cluster/operations/credhub-path-prefix.yml
- https://github.com/concourse/concourse-bosh-deployment/blob/master/cluster/operations/credhub.yml

**Blocked by this PR**: https://github.com/concourse/prod/pull/25
- we need the vars file to be updated first before this PR can be merged in
  - https://github.com/concourse/prod/pull/25/files#diff-86d753de7cb86f67ef2ba885e0bffd07


taylorsilva:
Closing because we're not doing this anymore.

PR #154, opened 2019-09-09
use limit-active-tasks in prod

nader-ziada:
Switch to `limit-active-tasks` in prod with a max-active-tasks-per-worker of 5 to begin with.

This PR should only be merged after https://github.com/concourse/concourse-bosh-deployment/pull/181

Signed-off-by: Nader Ziada <nziada@pivotal.io>


ddadlani:
fixes #4388

Issue #156, opened 2019-09-10
PR testing jobs should automatically pass on release-note-only PRs


pivotal-jamie-klassen:
it is very inconvenient to be forced to wait ~20 minutes to merge a documentation PR, especially when they are frequently release blockers!

PR #157, opened 2019-09-10
make documentation-only PRs pass faster

pivotal-jamie-klassen:
fixes concourse/ci#156

### DIFF
--- a/tasks/release-notes-only.yml
+++ b/tasks/release-notes-only.yml
@@ -1,0 +1,16 @@
+---
+platform: linux
+
+image_resource:
+  type: registry-image
+  source: {repository: concourse/unit}
+
+inputs:
+- name: concourse-pr
+
+run:
+  path: grep
+  args:
+  - -v
+  - release-notes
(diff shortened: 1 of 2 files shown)


tasks/release-notes-only.yml
```diff
@@ -0,0 +1,16 @@
...
+
+inputs:
+- name: concourse-pr
+
+run:
+  path: grep
+  args:
+  - -v
+  - release-notes
+  - concourse-pr/.git/resource/changed_files
```
julia-pu: Just to make sure, if new documentation files are added or deleted, they're also included as changed files?

pivotal-jamie-klassen: I'm pretty sure that's how it works - instinctively I'd expect the `changed_files` file to have something like the output of `git diff master --name-only --pretty=''` based on what we see [here](https://github.com/telia-oss/github-pr-resource#get). We haven't actually tested this, because it didn't seem worth the overhead of setting up a fake repo and fake PRs.

PR #162, opened 2019-09-10
Add "CONCOURSE_GARDEN_NETWORK_POOL" as supported BOSH config

deniseyu:
Improve error message if discrepancy between packaging and binary vars

Signed-off-by: Denise Yu <dyu@pivotal.io>


tasks/scripts/check-distribution-env/diff
```diff
@@ -28,6 +28,7 @@ main() {
     exit 0
   fi
 
+	echo "Error: found some variables in BOSH packaging that might not be supported by the Concourse binary:"
```
pivotal-jamie-klassen: Since this script is generic across bosh and helm, I think
```suggestion
	echo "Error: found some variables in ${DISTRIBUTION} packaging that might not be supported by the Concourse binary:"
```
would be better here.

deniseyu: makes sense!

PR #164, opened 2019-09-10
pass release-notes-only changes even faster

pivotal-jamie-klassen:
no need to spin up and tear down docker if you're not testing any code!


xtremerui:
Notice in PR pipeline, `watsjs` and `testflight` jobs run `yarn-build` still, is there a way we could also skip this if it is doc only changes?

pivotal-jamie-klassen:
@xtremerui #172 done.

PR #169, opened 2019-09-13
remove lidar from prod

pivotal-jwinters:
- Going to figure out a way to make it play nicer with the old scheduler


pivotal-jwinters:
RIP

PR #187, opened 2019-10-22
pre-5.5.4 concessions

pivotal-jamie-klassen:
mainly, fix the builder image, but also don't gate on k8s-topgun while it is
so flaky.

### DIFF
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -1235,7 +1235,7 @@ resources:
 - name: builder
   type: registry-image
   icon: *image-icon
-  source: {repository: concourse/oci-build-task}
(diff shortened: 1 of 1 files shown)


pipelines/release.yml
```diff
@@ -1237,7 +1235,7 @@ resources:
 - name: builder
   type: registry-image
   icon: *image-icon
-  source: {repository: concourse/builder-task}
+  source: {repository: concourse/oci-build-task}
```
vito:
```suggestion
  source: {repository: vito/oci-build-task}
```

pivotal-jamie-klassen: nice catch

PR #191, opened 2019-10-24
add `k8s-topgun` tag and `IN_CLUSTER` param to k8s-topgun

cirocosta:
After https://github.com/concourse/concourse/pull/4643, we're not able
to leverage the `IN_CLUSTER` environment variable, which switches the
behavior of using `port-forward` to reach endpoints by actually just
using the internal addresses of those.

Tagging the steps as `k8s-topgun` lets us be inside the cluster, making
`IN_CLUSTER` possible.

---

~update: we're finishing this off by doing acceptance with `fly execute` - will remove `wip:` afterwards~ ready!

### DIFF
--- a/tasks/scripts/k8s-topgun
+++ b/tasks/scripts/k8s-topgun
@@ -24,4 +24,4 @@ go mod download
 
 go install github.com/onsi/ginkgo/ginkgo
 
-ginkgo -nodes=4 -race -keepGoing -slowSpecThreshold=300 -skip="$SKIP" ./topgun/k8s/ "$@"
+ginkgo -nodes=8 -race -keepGoing -slowSpecThreshold=900 -flakeAttempts=3 -skip="$SKIP" ./topgun/k8s/ "$@"
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -464,27 +464,36 @@ jobs:
   - in_parallel:
     - get: concourse
       passed: [k8s-smoke]
+      tags: [k8s-topgun]
       trigger: true
     - get: version
       passed: [k8s-smoke]
+      tags: [k8s-topgun]
       trigger: true
     - get: concourse-rc-image
-      passed: [k8s-smoke]
-      trigger: true
       params: {format: oci}
+      passed: [k8s-smoke]
+      tags: [k8s-topgun]
+      trigger: true
     - get: concourse-rc-image-ubuntu
-      passed: [k8s-smoke]
-      trigger: true
       params: {format: oci}
-    - get: unit-image
-    - get: charts
-      trigger: true
       passed: [k8s-smoke]
+      tags: [k8s-topgun]
+      trigger: true
+    - get: unit-image
+      tags: [k8s-topgun]
+    - get: charts
+      passed: [k8s-smoke]
+      tags: [k8s-topgun]
+      trigger: true
     - get: ci
+      tags: [k8s-topgun]
   - task: k8s-topgun
     file: ci/tasks/k8s-topgun.yml
+    tags: [k8s-topgun]
     image: unit-image
     params:
+      IN_CLUSTER: "true"
       KUBE_CONFIG: ((kube_config))
       CONCOURSE_IMAGE_NAME: concourse/concourse-rc
   on_success: *fixed-concourse


tasks/scripts/k8s-topgun
```diff
@@ -24,4 +24,4 @@ go mod download
 
 go install github.com/onsi/ginkgo/ginkgo
 
-ginkgo -nodes=4 -race -keepGoing -slowSpecThreshold=300 -skip="$SKIP" ./topgun/k8s/ "$@"
+ginkgo -nodes=8 -race -keepGoing -slowSpecThreshold=900 -flakeAttempts=3 -flake-skip="$SKIP" ./topgun/k8s/ "$@"
```
cirocosta: with the more conservative use of `resources.requests` (see the PR that makes this one possible), we can bump the number of concurrent nodes here as we're putting less work per node.

cirocosta: ps.: `flakeAttempts`: https://github.com/onsi/ginkgo/pull/261

tasks/scripts/k8s-topgun
```diff
@@ -24,4 +24,4 @@ go mod download
 
 go install github.com/onsi/ginkgo/ginkgo
 
-ginkgo -nodes=4 -race -keepGoing -slowSpecThreshold=300 -skip="$SKIP" ./topgun/k8s/ "$@"
+ginkgo -nodes=8 -race -keepGoing -slowSpecThreshold=900 -flakeAttempts=3 -skip="$SKIP" ./topgun/k8s/ "$@"
```
cirocosta: `-flakeAttempts`: https://github.com/onsi/ginkgo/pull/261

`-nodes=8` (double what we used before) is not possible due to the use of more conservative `resources.requests` (see PR that made this possible).

cirocosta:
waaait wait wait

taylorsilva:
@cirocosta I saw wip :)

cirocosta:
we were having trouble with the `k8s-topgun` worker initializing containers, it turns out that the worker had landed in an ubuntu-based node 😅 

https://github.com/concourse/concourse/blob/addd0771777aed03f1a99c29f913edcd1ecf4ffa/topgun/k8s/container_limits_test.go#L29-L32

now we update the `nodeSelector` to make it always land on COS.

cirocosta:
Note.: we'll also have to get these changes back to the pipelines generated from `release.yml` (5.2.x, 5.5.x, etc).

We can do this in a separate PR after validating that it all works out.

Wdyt @pivotal-jamie-klassen ?

thx!

cirocosta:
yaaay acceptance worked!

https://ci.concourse-ci.org/builds/7128079

tasks/scripts/k8s-topgun
jomsie: flakeAttempts has a lively conversation around it onsi/ginkgo#260!

i do agree with some of the ideas expressed in that issue, would `flakeAttempts` be necessary with running topgun inside the k8s cluster? just wondering how stable k8s-topgun gets with the new setup...

jomsie:
also we've seen a bunch of [portforwarding issues with k8s-smoke](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/k8s-smoke/builds/3515#L5daf7bb3:20:31), is there any benefit to running k8s-smoke in the k8s cluster too? (sounds like a new issue for sure :D )

tasks/scripts/k8s-topgun
cirocosta: running it dozens of times showed that one time or another, there's a flake still 😬  I'd say we can keep track of that with a story around capturing those results like they said in the `ginkgo` issue (to keep track of the failures)

wdyt?

cirocosta: oh, and there's also the fact that we optimistically wait autoscaling to trigger on time, which can flake too

jomsie: you've thought about this, excellent! a story about capturing those results somehow would be pretty nifty. i'm convinced!

cirocosta:
@jomsie, yeah, I think so! 

I'd say that'd live in a different PR though, just so that we can isolate the changes 😁  

thanks!

jomsie:
this one LGT me!

PR #201, opened 2019-11-08
[fix concourse-ops#117] Concourse chart repo build pipeline

pivotal-bin-ju:
add a pipeline for building the concourse repo that is hosted on gcs.

- This pipeline updates the repo (bucket `concourse-chart`) against the chart source code https://github.com/concourse/concourse-chart.
- variable `concourse_chart_bucket_json_key` should be set before setup the pipeline.
- A workaround for chart name due to helm v2 require the chart name to be identical to the folder name. This is fixed in helm v3.


taylorsilva:
Going to move around the stuff done on this branch. Changing the release strategy a bit.

PR #200, opened 2019-11-08
Migrate to the new chart repo

pivotal-bin-ju:
fixes https://github.com/pivotal/concourse-ops/issues/117

We moved our helm chart out of the `helm/charts` repo and over to
concourse/concourse-chart

To make releasing and testing easier for us we're going to follow a
branching strategy similar to the concourse-bosh-deployment repo.

For the main concourse pipeline all tests will be done against a `dev`
branch that stays updated with master and has other commits that are
for unreleased changes currently on the master branch of
concourse/concourse

To reclarify, concourse/concourse-chart has two branches:
- `master` reflects the current OSS release of the chart
- `dev` reflects `master` plus other commits related to unreleased
versions of `concourse/concourse`

Currently this branching stragegy has no automation:
- `dev` not being promoted to master
- `dev` not being rebased on master before testing
- need to add lint checking for the pr against our helm chart

Related PR: https://github.com/concourse/concourse/pull/4735

cc @taylorsilva

PR #202, opened 2019-11-13
Release branch improvements

deniseyu:
Get rid of minor versions we don't support anymore in the `reconfigure` pipeline-setting logic, and use branches rather than master commit SHAs for concourse/concourse-charts.

### DIFF
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -2,7 +2,6 @@
 #
 #   ((release_major))                   the MAJOR version, e.g. 5
 #   ((release_minor))                   the MAJOR.MINOR version, e.g. 5.1
-#   ((charts_version))                  the commit on `helm/charts` where the version of
 #                                       concourse matches the desired release version
 #   ((concourse_smoke_deployment_name)) a unique name for the smoke bosh deployment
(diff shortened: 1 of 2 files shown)


pipelines/release.yml
```diff
@@ -1214,10 +1212,9 @@ resources:
 - name: charts
   type: git
   icon: *git-icon
-  version: {ref: ((charts_version))}
   source:
-    uri: https://github.com/helm/charts.git
-    branch: master
+    uri: https://github.com/concourse/concourse-chart.git
```
taylorsilva: This change may break k8s-topgun and anything else that used this chart repo. We hard-coded a bunch of `stable/concourse` stuff. Let me see where those transition things are at with the repo. 

Then again, this would be a really nice way of telling us where we need to make changes 😄

taylorsilva: Can we remove this diff from this PR? We have #200 which will make this change and also fix all the scripts this impacts.

deniseyu: It should be a no-op regardless of merge order I think?

taylorsilva: Guess so, no harm then! Will only break k8s stuff anyways which isn't a blocker for releasing.

PR #204, opened 2019-11-14
wip: dockerfiles/dev: add `containerd`

cirocosta:
coming along with https://github.com/concourse/concourse/pull/4752, this
commit adds `containerd` to our base (`concourse/dev`) so that we can
leverage `containerd` in `concourse/concourse:local` and other places in
the pipeline.

note that differently from `gdn`, we're not taking `containerd`, `runc`
and `cni` artifacts from the filesystem (populated by a pipeline) -
that's because there's quite a bit of version interdependency between
those that having them.

### DIFF
--- a/dockerfiles/dev/Dockerfile
+++ b/dockerfiles/dev/Dockerfile
@@ -10,6 +10,7 @@ FROM concourse/golang-builder
 
 RUN apt-get update && apt-get -y install \
       iproute2 \
+      curl \
       ca-certificates \
       file \
       btrfs-tools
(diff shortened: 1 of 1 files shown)


dockerfiles/dev/Dockerfile
```diff
@@ -21,7 +22,20 @@ RUN chmod +x /usr/local/bin/dumb-init
...
 RUN chmod +x /usr/local/concourse/bin/gdn
 
-# add fly executables
+# containerd executables
+ARG RUNC_VERSION=v1.0.0-rc8
+ARG CNI_VERSION=v0.8.2
+ARG CONTAINERD_VERSION=1.3.0
+
+RUN curl -sSL https://github.com/containerd/containerd/releases/download/v$CONTAINERD_VERSION/containerd-$CONTAINERD_VERSION.linux-amd64.tar.gz \
+      | tar -zvxf - -C /usr/local/concourse/bin --strip-components=1
```
cirocosta: this might look wrong (in the sense that we're not `mkdir -p /usr/local/concourse/bin` before), but that's because the previous step (`COPY gdn ...`) already creates that tree.

cirocosta:
> note that differently from gdn, we're not taking containerd, runc
and cni artifacts from the filesystem (populated by a pipeline) -
that's because there's quite a bit of version interdependency between
those that having them.

well, that does not necessarily mean that we have to hardcode URLs here though - we could still pin versions in the pipeline and achieve the same 🤔

cirocosta:
actually, those should come from a resource later on to form the final tarball anyway 😅  
Screen Shot 2019-11-14 at 2 20 24 PM

cirocosta:
related: https://github.com/concourse/concourse/issues/4993

cirocosta:
(superseded by the other PRs)

PR #207, opened 2019-11-14
track base resource pipelines in reconfigure.yml

pivotal-jamie-klassen:
contributes to concourse/prod#36


tasks/render-resource-pipeline-templates.yml
```diff
@@ -0,0 +1,13 @@
...
+  source: {repository: bitnami/jsonnet}
+inputs:
+- name: pipelines
+outputs:
+- name: rendered_pipelines
+params:
+  RESOURCES:
+run:
+  path: pipelines/tasks/scripts/render-resource-pipeline-templates
+  user: root
```
cirocosta: why do we need to explicitly tell that it's root? is it because `bitnami/jsonnet` has a different user set by default that concourse doesn't detect?

pipelines/reconfigure.yml
```diff
@@ -38,6 +38,82 @@ resources:
...
+        tracker
+        hg
+        github-release
+        semver
+        s3
+        concourse-pipeline
+        pool
+        datadog-event
+        mock
+  - put : prod
```
cirocosta: If I got it right, I think we don't need any other inputs other than `rendered_pipelines` 🤔  

```suggestion
  - put: prod
    inputs: [rendered_pipelines]
```

(soon we'll be able to use `detect` though 👀  https://github.com/concourse/concourse/pull/4616)

pivotal-jamie-klassen: hahaha I originally had it this way and opted for less code! Maybe it's worth just getting in the habit

tasks/render-resource-pipeline-templates.yml
pivotal-jamie-klassen: Something wacky about that image, yeah: https://hub.docker.com/r/bitnami/jsonnet/dockerfile the `USER 1001` doesn't have permission to read from the task input or write to the task output. Might be worth using a different image, I just liked the sales pitch.

tasks/scripts/render-resource-pipeline-templates
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e -u
+
+for resource in $RESOURCES; do
+  echo "configuring $resource..."
```
vito:
```suggestion
  echo "rendering '$resource' pipeline config..."
```

PR #208, opened 2019-11-15
Tag smoke-deployment resource to run on the topgun worker

taylorsilva:
Missed this change when making https://github.com/concourse/ci/pull/205

Without this, the resource check fails. Doesn't functionally affect the pipeline or ability to release.


pivotal-jamie-klassen:
I'm working on another PR and I want to merge this first to avoid having to do multiple rebases

PR #218, opened 2019-11-22
Removing the concourse-chart pipeline from reconfigure

taylorsilva:
This PR removes the `concourse-chart` pipeline from the `reconfigure` pipeline.
Don't need this old chart infrastructure since we're
maintaining the chart in our own repo now.

### DIFF
--- a/pipelines/reconfigure.yml
+++ b/pipelines/reconfigure.yml
@@ -13,20 +13,6 @@ resources:
     paths:
     - pipelines
 
-- name: charts-maintenance
-  type: git
-  icon: github-circle
-  source:
-    uri: https://github.com/concourse/charts
-    branch: maintenance
-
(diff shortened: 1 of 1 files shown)


pipelines/reconfigure.yml
```diff
@@ -13,20 +13,6 @@ resources:
...
     - pipelines
 
-- name: charts-maintenance
-  type: git
-  icon: github-circle
-  source:
-    uri: https://github.com/concourse/charts
-    branch: maintenance
-
-- name: maintenance-image
```
cirocosta: is this intended? 🤔 

I think both resources are still being used

taylorsilva:
I'm just going to close this, we're working on this as well as helm-prs in another track now.

PR #219, opened 2019-11-25
Revert "get release notes for patches from release branch"

pivotal-jwinters:
So master will be the only source of truth of all release notes. In individual release branch of `concourse` there is no need to keep release notes up to date.

### DIFF
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -1014,7 +1014,7 @@ resources:
   icon: *git-icon
   source:
     uri: https://github.com/concourse/concourse.git
-    branch: release/((release_minor)).x
+    branch: master
     paths:
     - release-notes/
 


vito:
I was chatting with @deniseyu and @cirocosta the other day, and I thought we wanted patch release notes to be on the release branches. Having them all on master means it takes two separate PRs for a single patch and means you can't include the release notes in PRs to a release branch. Is there something fundamentally broken with this approach?

pivotal-jwinters:
@vito we've noticed that there is no longer a source of truth for our release notes. They end up getting littered all over our release branches.

The other problem is that if you don't submit a PR with release notes the first time around, then you'll need the `concourse` resource to pass all the way through the pipeline again just to update release notes.

pivotal-jwinters:
Wait, nevermind that second bit isn't even true. The docs get threaded through as a separate resource.

pivotal-jwinters:
> Having them all on master means it takes two separate PRs for a single patch and means you can't include the release notes in PRs to a release branch

I think that point is more important than keeping a single source of truth so closing.

pivotal-jamie-klassen:
🤣 I'm starting to feel like there is no right answer to this. @deniseyu also suggested having some automation to merge the release notes back to master. When I think about where I would look for a summary of release notes, it would be cool to look on master. When I'm thinking about where to put release notes when I'm getting ready to ship a release, seeing the same (or worse, slightly different) set of release notes on master and my release branch might be pretty confusing.

PR #225, opened 2019-11-27
fix helm charts uri

### LINKED ISSUE #222: k8s-topgun needs helm/stable/postgresql and helm/table/prometheus
With the changes done in https://github.com/concourse/ci/commit/30e5721a20ad7574091ae0e6691704f79cf9642c, now `helm` contains only concourse-chart.

However in `k8s-topgun` job there are tests relying on [postgresql-chart](https://github.com/concourse/concourse/blob/db19f87d078278c32e71f962c0b32d0ceb2227d6/topgun/k8s/external_postgres_test.go#L18) and [prometheus-chart](https://github.com/concourse/concourse/blob/db19f87d078278c32e71f962c0b32d0ceb2227d6/topgun/k8s/prometheus_test.go#L29), which is now [broken](https://nci.concourse-ci.org/teams/main/pipelines/release-5.5.x/jobs/k8s-topgun/builds/5) in release/5.5.x pipeline, for example.

Do we need also `concourse/postgresql-chart` and `concourse/prometheus-chart` or we pull down those from `github.com/helm/charts` still?

xtremerui:
fix #222

### DIFF
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -1228,8 +1228,8 @@ resources:
   type: git
   icon: *git-icon
   source:
-    uri: https://github.com/concourse/concourse-chart.git
-    branch: release/((release_minor)).x
+    uri: https://github.com/helm/charts
+    branch: master
 
 - name: concourse-github-release
   type: github-release


pipelines/release.yml
```diff
@@ -1228,8 +1228,8 @@ resources:
   type: git
   icon: *git-icon
   source:
-    uri: https://github.com/concourse/concourse-chart.git
-    branch: release/((release_minor)).x
+    uri: https://github.com/helm/charts
+    branch: master
```
cirocosta: @taylorsilva heey, wdyt of this approach? (context here: https://github.com/concourse/ci/issues/222#issuecomment-559174540)

taylorsilva: This approach makes sense 👍

PR #231, opened 2019-12-02
Add deploy step for bosh-topgun worker

taylorsilva:
fixes https://github.com/concourse/prod/issues/43

This PR adds a `put` step to redeploy our bosh topgun worker.

This `put` step, added to when we redeploy ci, ensures that our bosh
external worker is updated at the same time. Had to add `worker_private_key` and `tsa_host_public_key`, at the pipeline level in vault (`concourse/main/concourse/`)

### DIFF
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -1104,6 +1104,28 @@ jobs:
     - get: bbr-sdk-release
     - get: prod
     - get: ci
+  - put: bosh-topgun-worker-deployment
+    tags: [topgun]
+    params:
+      manifest: cbd/cluster/external-worker.yml
+      stemcells:
+      - gcp-xenial-stemcell/*.tgz
+      releases:
+      - concourse-release/*.tgz
+      - bpm-release/*.tgz
+      ops_files:
+      vars_files:
+      - cbd/versions.yml
+      vars:
+        deployment_name: topgun-worker
+        instances: 1
+        azs: [z1]
+        worker_vm_type: worker
+        external_worker_network_name: default
+        worker_tags: [topgun]
+        tsa_host: "nci.concourse-ci.org"
+        tsa_host_key.public_key: ((tsa_host_public_key))
+        worker_key: ((worker_private_key))
   - put: prod-deployment
     params:
       manifest: cbd/cluster/concourse.yml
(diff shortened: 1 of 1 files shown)


pipelines/concourse.yml
```diff
@@ -1104,6 +1104,27 @@ jobs:
     - get: bbr-sdk-release
     - get: prod
     - get: ci
+  - put: bosh-topgun-worker-deployment
```
pivotal-jamie-klassen: I think this step should have `tags: [topgun]` based on https://concourse-ci.org/tags-step-modifier.html#tags

pipelines/concourse.yml
```diff
@@ -1838,6 +1859,16 @@ resources:
     client_secret: ((bosh_client.secret))
     deployment: concourse-prod
 
+- name: bosh-topgun-worker-deployment
```
pivotal-jamie-klassen: I also think this resource should have `tags: [topgun]` https://concourse-ci.org/resources.html#resource-tags

taylorsilva:
@pivotal-jamie-klassen you raise a great point and I believe you are correct! Added that tag, review again please?

pipelines/concourse.yml
```diff
@@ -1104,6 +1104,28 @@ jobs:
...
+  - put: bosh-topgun-worker-deployment
+    tags: [topgun]
+    params:
+      manifest: cbd/cluster/external-worker.yml
+      stemcells:
+      - gcp-xenial-stemcell/*.tgz
+      releases:
+      - concourse-release/*.tgz
+      - bpm-release/*.tgz
+      ops_files:
```
pivotal-jamie-klassen: I suspect that the `bosh-deployment-resource` is gonna expect this to be an array. Can you drop the param completely?

pipelines/concourse.yml
```diff
@@ -1104,6 +1104,28 @@ jobs:
...
+      vars_files:
+      - cbd/versions.yml
```
pivotal-jamie-klassen: I think we're gonna want to deploy RCs, so can we maybe drop this and just add the relevant values below? In fact, the default for `concourse_version` at least might be `latest`, which would be what we want in this case.

pipelines/concourse.yml
taylorsilva: Yup, that was a mistake even having it there.

pipelines/concourse.yml
taylorsilva: Added an ops-file that switches the versions over to `latest`, `dev-versions.yml`

taylorsilva:
Made the changes requested by @pivotal-jamie-klassen. Now using an ops-file that will allow us to use rc versions

PR #237, opened 2019-12-09
scale down old prod with only one worker and web instance

jamieklassen:
Fixes: concourse/prod#46

- remove `bosh-prod-deploy` job from `ci`
- remove the papertrail syslog


pipelines/concourse.yml
```diff
@@ -1071,131 +1070,6 @@ jobs:
   on_success: *fixed-concourse
   on_failure: *broke-concourse
 
-- name: bosh-prod-deploy
```
taylorsilva: If you rebase on master now, you'll see that removing this would also mean removing the new `put` step that re-deploys the topgun-worker we have on BOSH. 

https://github.com/concourse/ci/blob/43d09072d69eeecafd547596a0e608bc9941e184/pipelines/concourse.yml#L1107-L1127

Could we rename this job to `prod-deploy` and have it be the job where we re-deploy prod and all it's external workers? I'm not saying do that work in this PR, but setup the home for these future tasks that we know we need to do somewhere.

vito:
Let's just make sure prod is deleted.

xtreme-sameer-vohra:
Prod is still up. Gave notice to folks of its impending doom incase anyone needs it. 
Will delete it end of day today if no one objects.

xtreme-sameer-vohra:
Scaling down prod first pass
```
$ bosh -e prod -d concourse-prod deploy /tmp/prod.yml
Using environment 'https://bosh.concourse-ci.org' as user 'admin'

Using deployment 'concourse-prod'

Release 'windows-utilities/0.11.0' already exists.

  instance_groups:
  - name: web
-   instances: 3
+   instances: 0
  - name: worker
-   instances: 8
+   instances: 0
  - name: untrusted-worker
-   instances: 3
+   instances: 0
  - name: windows-worker
-   instances: 1
+   instances: 0

Continue? [yN]: y

Task 1541841
```

The only VMs that will remain will be vault and postgres which are stateful. Will delete them at the end of the week.

xtreme-sameer-vohra:
Deleted the `prod` deployment from the `prod` director.

vito:
Guess this can be closed then?

PR #240, opened 2019-12-10
use `https` in `bin-smoke`

pivotal-jamie-klassen:
Added tests for concourse/concourse#4795
Hopefully concourse/concourse#4869 will make these tests pass

### DIFF
--- a/deployments/smoke/systemd/smoke-web.conf.tpl
+++ b/deployments/smoke/systemd/smoke-web.conf.tpl
@@ -1,8 +1,11 @@
 [Service]
-Environment=CONCOURSE_EXTERNAL_URL=http://${instance_ip}:8080
+AmbientCapabilities=CAP_NET_BIND_SERVICE
+Environment=CONCOURSE_EXTERNAL_URL=${instance_url}
 Environment=CONCOURSE_SESSION_SIGNING_KEY=/etc/concourse/session_signing_key
 Environment=CONCOURSE_TSA_HOST_KEY=/etc/concourse/host_key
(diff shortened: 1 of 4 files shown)


deployments/smoke/systemd/smoke-web.conf.tpl
```diff
@@ -1,8 +1,11 @@
 [Service]
-Environment=CONCOURSE_EXTERNAL_URL=http://${instance_ip}:8080
+AmbientCapabilities=CAP_NET_BIND_SERVICE
+Environment=CONCOURSE_EXTERNAL_URL=${instance_url}
 Environment=CONCOURSE_SESSION_SIGNING_KEY=/etc/concourse/session_signing_key
 Environment=CONCOURSE_TSA_HOST_KEY=/etc/concourse/host_key
 Environment=CONCOURSE_TSA_AUTHORIZED_KEYS=/etc/concourse/authorized_worker_keys
+Environment=CONCOURSE_TLS_BIND_PORT="443"
+Environment=CONCOURSE_ENABLE_LETS_ENCRYPT="True"
```
vito:
```suggestion
Environment=CONCOURSE_ENABLE_LETS_ENCRYPT="true"
```

...in the (possibly likely?) event that the implementation for bool env vars is something like `v == "true"`.

vito: Also: we should make sure to use their [staging environment](https://letsencrypt.org/docs/staging-environment/), which has much higher rate-limits as it's meant for testing.

URL is https://acme-staging-v02.api.letsencrypt.org/directory

PR #246, opened 2019-12-17
add integration-image for integration test in testflight

kcmannem:
fixes concourse/concourse#4878

Signed-off-by: Bin Ju <bju@pivotal.io>
Co-authored-by: Krishna Mannem <kmannem@pivotal.io>

### DIFF
--- a/dockerfiles/integration/Dockerfile
+++ b/dockerfiles/integration/Dockerfile
@@ -1,0 +1,31 @@
+# This Dockerfile is only for running integration tests that depend on a running containerd process
+# Not intended to be packaged or used for production
+
+#FROM concourse/golang-builder
+FROM concourse/unit
+
+RUN mkdir -p /go/src/github.com/concourse/ /usr/local/concourse/bin/
+ENV PATH=/usr/local/concourse/bin:$PATH
+ENV GO111MODULE=on
+
+COPY start.sh /usr/local/concourse/bin/start.sh
+VOLUME /go/src/github.com/concourse/concourse
+VOLUME /go
+
+ARG RUNC_VERSION=v1.0.0-rc8
+ARG CNI_VERSION=v0.8.2
+ARG CONTAINERD_VERSION=1.3.0
+
+RUN apt update && apt install -y curl vim
+
+RUN curl -sSL https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 -o /usr/local/concourse/bin/init && chmod +x /usr/local/concourse/bin/init
+RUN set -x && \
+	curl -sSL https://github.com/containerd/containerd/releases/download/v$CONTAINERD_VERSION/containerd-$CONTAINERD_VERSION.linux-amd64.tar.gz \
+		| tar -zvxf - -C /usr/local/concourse/bin --strip-components=1 && \
+	curl -sSL https://github.com/opencontainers/runc/releases/download/$RUNC_VERSION/runc.amd64 \
+		-o /usr/local/concourse/bin/runc && chmod +x /usr/local/concourse/bin/runc && \
+	curl -sSL https://github.com/containernetworking/plugins/releases/download/$CNI_VERSION/cni-plugins-linux-amd64-$CNI_VERSION.tgz \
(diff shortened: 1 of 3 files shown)


pipelines/concourse.yml
```diff
@@ -128,6 +128,30 @@ jobs:
         mode: normal
         alert_type: broke
 
+- name: build-integration-image
```
deniseyu: could we call this `build-containerd-integration-image` or something similar. Someone who rolls onto the team might find it extremely confusing that this isn't the image used for our main integration tests

cirocosta: if I understood correctly, it's not *only* for testing `containerd` - it'd be used for the whole set of testflight tests, *as well as* `containerd`'s integration ones.

is that right, @kcmannem?

thx!

dockerfiles/integration/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Not intended to be packaged or used for production
+
+#FROM concourse/golang-builder
+FROM concourse/unit
+
+RUN mkdir -p /go/src/github.com/concourse/ /usr/local/concourse/bin/
+ENV PATH=/usr/local/concourse/bin:$PATH
+ENV GO111MODULE=on
+
+COPY start.sh /usr/local/concourse/bin/start.sh
```
cirocosta: where is this `start.sh` coming from? 🤔  

```
 ~/workspace/ci $ git status
On branch containerd-integration
Your branch is up to date with 'origin/containerd-integration'.

nothing to commit, working tree clean
 ~/workspace/ci $ find . -name "start.sh"
```

dockerfiles/integration/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# This Dockerfile is only for running integration tests that depend on a running containerd process
```
cirocosta: nit: from what I understood, it's supposed to not only running `integration` tests, but also `testflight`. 

```suggestion
 # This Dockerfile is only for running testflight and other integration tests
```

dockerfiles/integration/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Not intended to be packaged or used for production
+
+#FROM concourse/golang-builder
```
cirocosta: nit: leftovers? 😁

pipelines/concourse.yml
pivotal-bin-ju: @cirocosta you're right! It is the easiest way we could run both existed test flights and `containerd` together maybe. Any concern on this?

pivotal-bin-ju: @deniseyu good point! But I am worrying whether it is going to confuse people after renaming it to that name, while it is used for both for the whole test flights.
What about `build-containerd-enabled-integration-image`?

cirocosta: > @cirocosta [...]  Any concern on this?

ooh, no concerns! I was just raising the "why" I think this was not named after `*-containerd-*` 😁

cirocosta:
Hey @pivotal-bin-ju @kcmannem , I just updated the pipeline so that we can `fly set-pipeline` it.

Could take a quick look at the comments above though? Specifically with regards to `start.sh`.

Thanks!

dockerfiles/integration/Dockerfile
cirocosta:
```suggestion

```

cirocosta:
It got me wondering whether we really need to have a separate image for this - we use `unit` (which sure, could be renamed to something better) for packing all of the possibly necessary pieces of extra software needed for testing in various conditions, so, why not include containerd there? we already include `kubectl`, `helm`, etc ...

 I think the benefit of having it all in the same place is that it just becomes simpler to keep track of where those dependencies are - have a problem with a dep? just go straight to `unit`.

wdyt of having it there?

thanks!

cirocosta:
Hey,

I went forward with the changes I described in https://github.com/concourse/ci/pull/246#issuecomment-571782505, building upon the contributions here, so I'll go ahead w/ closing this one.

Thank you all!

PR #247, opened 2019-12-17
Rename references to the testing bosh director

zoetian:
Fixes: pivotal/concourse-ops#176

We noticed that the `topgun` worker tags would be used on both `smoke` and `topgun` test, so it's better to rename it as `bosh`. Also moved the vault secrets for the testing bosh director.
```
topgun_ca_cert
topgun_bosh_client
```
to 
```
testing_bosh_ca_cert
testing_bosh_client
```

The `bosh-smoke` job in the main pipeline now uses the testing bosh director instead of the prod director.

### DIFF
--- a/pipelines/branch.yml
+++ b/pipelines/branch.yml
@@ -552,9 +552,9 @@ jobs:
     image: unit-image
     params:
       BOSH_ENVIRONMENT: https://10.0.0.6:25555
-      BOSH_CA_CERT: ((topgun_ca_cert))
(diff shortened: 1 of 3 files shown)


xtreme-sameer-vohra:
@zoetian @pivotal-jamie-klassen I believe the branch restrictions have been lifted off the ci repo, so we can merge directly.

Thanks for updating branch and release yamls as well 👍

PR #251, opened 2019-12-19
parameterized concourse_image tag

xtremerui:
now we are doing beta release pipeline for 6.0 in https://nci.concourse-ci.org/teams/main/pipelines/release-6.0.x

which is not published yet. So there isn't a valid tag `6.0.0` for `concourse/concourse` on dockerhub. We need to set it to `5.7.2`.

### DIFF
--- a/pipelines/reconfigure.yml
+++ b/pipelines/reconfigure.yml
@@ -189,6 +189,7 @@ jobs:
         vars:
           release_major: "5"
           release_minor: "5.2"
+          latest_release: "5.2"
           concourse_smoke_deployment_name: "concourse-smoke-5-2"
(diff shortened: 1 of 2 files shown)


pipelines/reconfigure.yml
```diff
@@ -219,6 +222,7 @@ jobs:
         vars:
           release_major: "6"
           release_minor: "6.0"
+          latest_release: "5.7.2"
```
pivotal-jamie-klassen:
```suggestion
          latest_release: "5"
```
might be more future-proof

xtremerui: Good call thx!

PR #252, opened 2020-01-06
Add pipeline to publish pivnet version of the chart

taylorsilva:


pipelines/reconfigure.yml
```diff
@@ -244,6 +255,12 @@ jobs:
         vars:
           branch_name: use_zstd
           environment: use-zstd
+      - name: enterprise-helm-chart
+        team: main
+        config_file: concourse-for-k8s/ci/pipeline.yml
+        vars:
```
cirocosta: Hey,

If I understood this correctly, there's a missing variable here that will not exist on vault:

```
 fly -t nci \
 	set-pipeline \
		--check-creds \
		-p test1 \
		-c pipeline.yml \
		--var=concourse_minor=5.5 \
		--var=concourse_chart_branch=release/5.5.x
```

which complains with:

```
1 error occurred:
	* failed to interpolate task config: undefined vars: concourse_version
```

which, makes sense 🤔 
```
      - task: create
        config:
          platform: linux
          params:
            VERSION: ((concourse_version))
          inputs:
```

https://github.com/pivotal/concourse-for-k8s/blob/2693012d04255d54901e1cc8865736ea38127acb/ci/pipeline.yml#L86-L91

taylorsilva: Great catch. I removed those lines from the pipeline yaml now. I'd prefer to track which version of Concourse the pipeline is using via git commits in the `concourse-for-k8s` repo instead of in the `ci` repo and re-setting the pipeline.

https://github.com/pivotal/concourse-for-k8s/blob/master/VERSION

cirocosta: cooool, lgtm!

PR #253, opened 2020-01-08
pipelines/concourse: add containerd-integration

cirocosta:
As the `containerd` backend integration tests require `containerd` to be
running, we need to have its binaries present in the image that forms
the base that runs such tests.

Given that `unit` has been the place where all extra dependencies get
added, there's where `containerd` end up being added to.

For the execution of the tests, a separate job was created so that we
don't get possible failures from `testflight` when the problem is
isolated to `containerd` only.

---

This is building upon https://github.com/concourse/ci/pull/246#issuecomment-571782505, where
I proposed to go with a separate job for the containerd integration, while
keeping `containerd` (the binary) itself in `unit`.

Screen Shot 2020-01-08 at 9 26 59 AM

Wdyt @concourse/runtime ? 

thanks! 

---

marked as `wip` until I get a full run of it through `fly execute`

### DIFF
--- a/dockerfiles/unit/Dockerfile
+++ b/dockerfiles/unit/Dockerfile
@@ -78,3 +78,19 @@ RUN curl -so /etc/ssl/certs/fake_root.pem \
     -t "TCu,Cu,Tu" \
     -i /etc/ssl/certs/fake_root.pem \
     -d sql:/root/.pki/nssdb
+
+# install containerd
+ARG RUNC_VERSION=v1.0.0-rc8
+ARG CNI_VERSION=v0.8.2
+ARG CONTAINERD_VERSION=1.3.0
+
+ENV PATH=$PATH:/usr/local/concourse/bin
+
+RUN set -x && \
+	mkdir -p /usr/local/concourse/bin && \
+	curl -sSL https://github.com/containerd/containerd/releases/download/v$CONTAINERD_VERSION/containerd-$CONTAINERD_VERSION.linux-amd64.tar.gz \
+		| tar -zvxf - -C /usr/local/concourse/bin --strip-components=1 && \
+	curl -sSL https://github.com/opencontainers/runc/releases/download/$RUNC_VERSION/runc.amd64 \
+		-o /usr/local/concourse/bin/runc && chmod +x /usr/local/concourse/bin/runc && \
+	curl -sSL https://github.com/containernetworking/plugins/releases/download/$CNI_VERSION/cni-plugins-linux-amd64-$CNI_VERSION.tgz \
+		| tar -zvxf - -C /usr/local/concourse/bin
--- a/dockerfiles/unit/Dockerfile
+++ b/dockerfiles/unit/Dockerfile
@@ -85,8 +85,6 @@ ARG CNI_VERSION=v0.8.3
 ARG CONTAINERD_VERSION=1.3.2
 ARG INIT_VERSION=1.2.2
 
-ENV PATH=$PATH:/usr/local/concourse/bin
-
 RUN set -x && \
 	mkdir -p /usr/local/concourse/bin && \
 	curl -sSL https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 \
--- a/dockerfiles/unit/Dockerfile
+++ b/dockerfiles/unit/Dockerfile
@@ -89,6 +89,8 @@ ENV PATH=$PATH:/usr/local/concourse/bin
 
 RUN set -x && \
 	mkdir -p /usr/local/concourse/bin && \
+	curl -sSL https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 \
+		-o /usr/local/concourse/bin/init && chmod +x /usr/local/concourse/bin/init && \
 	curl -sSL https://github.com/containerd/containerd/releases/download/v$CONTAINERD_VERSION/containerd-$CONTAINERD_VERSION.linux-amd64.tar.gz \
 		| tar -zvxf - -C /usr/local/concourse/bin --strip-components=1 && \
 	curl -sSL https://github.com/opencontainers/runc/releases/download/$RUNC_VERSION/runc.amd64 \
--- a/dockerfiles/unit/Dockerfile
+++ b/dockerfiles/unit/Dockerfile
@@ -80,9 +80,10 @@ RUN curl -so /etc/ssl/certs/fake_root.pem \
     -d sql:/root/.pki/nssdb
 
 # install containerd
-ARG RUNC_VERSION=v1.0.0-rc8
-ARG CNI_VERSION=v0.8.2
-ARG CONTAINERD_VERSION=1.3.0
+ARG RUNC_VERSION=v1.0.0-rc9
+ARG CNI_VERSION=v0.8.3
+ARG CONTAINERD_VERSION=1.3.2
+ARG INIT_VERSION=1.2.2
 
 ENV PATH=$PATH:/usr/local/concourse/bin
 
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -308,7 +308,7 @@ jobs:
 
 - name: containerd-integration
   public: true
-  max_in_flight: 2
+  max_in_flight: 1
   plan:
   - in_parallel:
     - get: concourse
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -21,6 +21,7 @@ groups:
   - unit
   - dev-image
   - testflight
+  - containerd-integration
   - watsjs
   - rc
   - build-rc
@@ -302,6 +303,24 @@ jobs:
   - put: dev-image
     params: {image: image/image.tar}
     get_params: {format: oci}
+  on_success: *fixed-concourse
+  on_failure: *broke-concourse
+
+- name: containerd-integration
+  public: true
+  max_in_flight: 2
+  plan:
+  - in_parallel:
+    - get: concourse
+      passed: [unit]
+      trigger: true
+    - get: unit-image
+    - get: ci
+  - task: integration
+    image: unit-image
+    privileged: true
+    timeout: 1h
+    file: ci/tasks/containerd-integration.yml
   on_success: *fixed-concourse
   on_failure: *broke-concourse
 
--- a/tasks/scripts/containerd-integration
+++ b/tasks/scripts/containerd-integration
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -e -u -x
+set -e -u
 
 export GOPATH=$PWD/gopath
 export PATH=$GOPATH/bin:$PATH
--- a/tasks/scripts/containerd-integration
+++ b/tasks/scripts/containerd-integration
@@ -34,7 +34,7 @@ state = "/run/containerd"
     shim = "containerd-shim"
 ' >./containerd.toml
 
-  containerd --config ./containerd.toml >/tmp/containerd.log 2>&1 &
+  PATH=/usr/local/concourse/bin containerd --config ./containerd.toml >/tmp/containerd.log 2>&1 &
   echo $! >/tmp/containerd.pid
 
   sleep 1
--- a/tasks/scripts/containerd-integration
+++ b/tasks/scripts/containerd-integration
@@ -17,7 +17,7 @@ start_containerd() {
   mkdir -p /scratch/containerd
 
   echo '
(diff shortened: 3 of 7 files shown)


tasks/scripts/containerd-integration
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -e -u -x
```
cirocosta: whops, I should get rid of that `-x`

tasks/scripts/containerd-integration
```diff
@@ -0,0 +1,37 @@
...
+
+source ci/tasks/scripts/docker-helpers.sh
+
+main() {
+  sanitize_cgroups
+  start_containerd
+  run_test
+}
+
+start_containerd() {
```
cirocosta: here I tried to follow as closely as possible what we already do for `docker`:

https://github.com/concourse/ci/blob/e196f9c67c3508ac50e3409e1af180c11aa960c0/tasks/scripts/docker-helpers.sh#L64-L72

dockerfiles/unit/Dockerfile
```diff
@@ -78,3 +78,19 @@ RUN curl -so /etc/ssl/certs/fake_root.pem \
     -t "TCu,Cu,Tu" \
     -i /etc/ssl/certs/fake_root.pem \
     -d sql:/root/.pki/nssdb
+
+# install containerd
+ARG RUNC_VERSION=v1.0.0-rc8
```
cirocosta: these are actually not the up to date versions - should be:

```
ARG RUNC_VERSION=v1.0.0-rc9
ARG CNI_VERSION=v0.8.3
ARG CONTAINERD_VERSION=1.3.2
```

dockerfiles/unit/Dockerfile
```diff
@@ -78,3 +78,19 @@ RUN curl -so /etc/ssl/certs/fake_root.pem \
...
+ARG CNI_VERSION=v0.8.2
+ARG CONTAINERD_VERSION=1.3.0
+
+ENV PATH=$PATH:/usr/local/concourse/bin
+
+RUN set -x && \
+	mkdir -p /usr/local/concourse/bin && \
```
cirocosta: missing the `init` that's meant to be used as the replacement for `/tmp/gdn-init`

tasks/scripts/containerd-integration
```diff
@@ -0,0 +1,37 @@
...
+  containerd >/tmp/containerd.log 2>&1 &
```
cirocosta: we should be providing our own configuration here to ensure that we don't get any collisions with `docker`'s `runc`, etc.

dockerfiles/unit/Dockerfile
cirocosta: https://github.com/concourse/ci/pull/253/commits/37f0c2608dd24113d0e43979b5f34dba70fc5c55

dockerfiles/unit/Dockerfile
cirocosta: https://github.com/concourse/ci/pull/253/commits/7e94b5e730b45bf3d0c7251f3ad56dd2f676191f

tasks/scripts/containerd-integration
cirocosta: https://github.com/concourse/ci/pull/253/commits/b6f26951005f1917e9ce22f0066586f2919afbfb

tasks/scripts/containerd-integration
cirocosta: https://github.com/concourse/ci/pull/253/commits/5bee7071942eef142bc15c7ae3eb2ef9b02038ec

pipelines/concourse.yml
```diff
@@ -305,6 +306,24 @@ jobs:
   on_success: *fixed-concourse
   on_failure: *broke-concourse
 
+- name: containerd-integration
+  public: true
+  max_in_flight: 2
```
pivotal-jamie-klassen: Is this a magic number?

tasks/scripts/containerd-integration
```diff
@@ -0,0 +1,57 @@
...
+
+[debug]
+  address = "/run/containerd/debug.sock"
+  level = "debug"
+
+[plugins]
+  [plugins."io.containerd.runtime.v1.linux"]
+    runtime = "runc"
+    shim = "containerd-shim"
+' >./containerd.toml
```
pivotal-jamie-klassen: is it maybe worth tracking this config in a separate file? I could see it going into `deployments/`, which also holds our terraform config for bin-smoke. I just realized that the README is wrong to describe it as containing BOSH-related configuration.

pipelines/concourse.yml
cirocosta: oh good catch! that was straight from `testflight` - while I can see the desire to limit the possible concurrent runs to not have N of them, I can't come up with a reason why 2 rather than, say, 3.

```
ag -Q "max_in_flight"
pipelines/concourse.yml
311:  max_in_flight: 2
329:  max_in_flight: 2
353:  max_in_flight: 2
564:  max_in_flight: 1
802:  max_in_flight: 1
```

I'll go with 1 per the desire to limit, but not have an opinion on "how much" 👍

tasks/scripts/containerd-integration
cirocosta: > is it maybe worth tracking this config in a separate file?

yeahh, I think it makes sense - just how for `smoke` we have the `systemd` service configuration there 👍  

>  I just realized that the README is wrong to describe it as containing BOSH-related configuration.

thaaat's indeed wrong  haaha

cirocosta:
The only thing holding me back at this point is the fact that the test execution still fails 😞 

I tried debugging this quite a bit, but I can't figure out why it works flawlessly from outside the container (using [`cirocosta/containerd-install`](https://github.com/cirocosta/containerd-install) to set up the system) but not from the task


```
=== RUN   TestSuite
=== RUN   TestSuite/TestContainerCreateAndDestroy
--- FAIL: TestSuite (0.04s)
    integration_test.go:54:
                Error Trace:    integration_test.go:54
                Error:          Received unexpected error:
                                client error: failed to create a task in container: io.containerd.runc.v2: failed to connect: dial unix: missing address
                                : exit status 1: unknown
                Test:           TestSuite
    --- FAIL: TestSuite/TestContainerCreateAndDestroy (0.03s)
        testing.go:864: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
FAIL

There were failures detected in the following suites:
        integration ./worker/backend/integration
``` 

I'm 90% sure it's related to some form of automatic feature detection.

---

If anyone's curious about how to debug this using `dlv` to place breakpoints etc in `containerd`'s code: you *have* to rebuild `containerd` - it seems like they're stripping their final binary.

So, to get around that:

1. make sure you get all of your dependencies right
  - that's pretty much `libseccomp-dev` and `libbtrfs-dev`

2. clone the repo (github.com/containerd/containerd), checkout to the tag that matches the version our backend implementation expects (1.3.2 at this moment)

3. build
  - plain `go build ./cmd/containerd`

4. run containerd

5. `dlv attach $(pidof containerd) /proc/$(pidof containerd)/exe` (w/ uid=0)

cirocosta:
Got a successful run! The problem was pretty much what I described in https://github.com/concourse/ci/pull/253/commits/47bdad1df57fad1816a057d1b2a0bda8b2e84a4a:

```


Previously, `unit` had PATH configured with higher priorities in the
search for containerd binaries to those paths that were not
`/usr/local/concourse/bin`. This was to that when `docker` comes up (or
would create a container), it'd pick the right binaries for its version.

By doing so, when running our own version of `containerd`, we'd end up
picking the binaries that suited `docker`'s version, not ours.

Thus, having our `containerd` starting with PATH set to the place where
we know our binaries exist, forces it to find the right binaries.
```

cc @pivotal-jamie-klassen

tasks/scripts/containerd-integration
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+set -e -u
+
+export GOPATH=$PWD/gopath
+export PATH=$GOPATH/bin:$PATH
+
+source ci/tasks/scripts/docker-helpers.sh
```
pivotal-jamie-klassen: hmmm, this file's name becomes misleading with this PR. If the same helper is used for containerd, it's not so much about docker so much as... allowing nested containers? honestly that `sanitize_cgroups` function has always gone way over my head.

tasks/scripts/containerd-integration
pivotal-jamie-klassen: so the other thing we do for docker is [make sure it stops](https://github.com/concourse/ci/blob/master/tasks/scripts/with-docker-compose#L31-L38) when the build is done -- since otherwise i think we can leak the `docker` process. I'm not totally sure how this black magic works, but should we maybe be careful about stopping `containerd`?

pipelines/concourse.yml
```diff
@@ -305,6 +306,24 @@ jobs:
   on_success: *fixed-concourse
   on_failure: *broke-concourse
 
+- name: containerd-integration
+  public: true
+  max_in_flight: 1
```
pivotal-jamie-klassen: just realized we had similar concurrency limits on `testflight` and `watsjs` - obviously they run in containers so they should be independent/reproducible - are they super expensive to run? maybe because they're privileged? i really don't have a solid grasp on this.

pivotal-jamie-klassen:
> For trying the task out, I'm currently going with:
> 
> ```
> fly -t local execute \
>   --privileged \
>   -c tasks/containerd-integration.yml \
>   -i=concourse=$(realpath ~/workspace/concourse) \
>   -i ci=.
> ```
> 
> having `concourse` being a local copy of `concourse/concourse`, the current dir being this PR, and with the task referencing a container image based of the modified `unit` image that this PR changes.  

I often find myself following workflows like this, and I feel they would benefit from a feature like https://github.com/concourse/concourse/issues/4981#issuecomment-572852925.

tasks/scripts/containerd-integration
cirocosta: That's a good point!

I think the big reason for stopping the whole thing there is that we run a
quite expensive setup - we're essentially running an entire Concourse
installation.

While I do not oppose cleaning things so that we end up with less resources
being left behind, I wonder if that's a pattern to persue or not even when the
resource utilization is not high, in the sense that I see as a nice to have the
ability to hijack and then play around w/ the tests having all set up already.

e.g., for *every* task where a daemon/process gets "background'ed", should we
always set up an exit trap to ensure that it gets killed?

I'm leaning towards "if it's super expensive, then, yeah", otherwise, "leave
there" as it can be convenient.

wdyt?

tasks/scripts/containerd-integration
cirocosta: > this file's name becomes misleading with this PR. [...] it's not so much about docker so much as... allowing nested containers?

oh totally! that's the sole purpose for `source`ing that file in

do you think it'd be better having a "cgroups-helpers.sh" (or a better name) that we consume in both `docker-helpers` & `containerd-integration`?

tasks/scripts/containerd-integration
pivotal-jamie-klassen: Yeah, we were definitely encountering cases where workers were getting overloaded by those zombie `dockerd`s. I'm guessing that `containerd`, as it's being exercised here, will not be quite as expensive?

tasks/scripts/containerd-integration
pivotal-jamie-klassen: Yeah, that sounds ideal! Also I'd just like to better understand what that function does haha

tasks/scripts/containerd-integration
cirocosta: > will not be quite as expensive?

yeahh, it'll be super cheap 😁  even if we leave containers "running", those containers will do nothing, as opposed to Concourse, which does all of its gc, scanning, scheduling, etc

tasks/scripts/containerd-integration
cirocosta: >  I'd just like to better understand what that function does haha

I just added some comments to make it clearer 😁  

see f14a660928e3624100ac65449869a83265b1ab38

cirocosta: btw, I went with the "singular form" (`cgroup`) rather than plural (`cgroups`):

>  "cgroup" stands for "control group" and is never capitalized.  **The
singular form is used to designate the whole feature and also as a
qualifier as in "cgroup controllers"**.  When explicitly referring to
multiple individual control groups, the plural form "cgroups" is used.

pivotal-jamie-klassen: That's awesome, thanks!

PR #256, opened 2020-01-17
dockerfiles/unit: bump nodejs version

cirocosta:
the version of nodejs that we're using just reached its maintenance end
of life.

despite v12 not being the very latest version of node, it's the latest
LTS, which  gives us maintenance up until 2022-04-30.

### DIFF
--- a/dockerfiles/unit/Dockerfile
+++ b/dockerfiles/unit/Dockerfile
@@ -18,7 +18,7 @@ RUN apt-get update && apt-get -y install postgresql-11
 ENV PATH $PATH:/usr/lib/postgresql/11/bin
 
 # install Node 8.x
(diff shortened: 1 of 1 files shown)


dockerfiles/unit/Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN apt-get update && apt-get -y install postgresql-11
 ENV PATH $PATH:/usr/lib/postgresql/11/bin
 
 # install Node 8.x
```
pivotal-jamie-klassen: felt hard to ignore
```suggestion
 # install Node 12.x
```

cirocosta: whoops 😁

PR #257, opened 2020-01-20
Make the alpine image actually based on alpine

taylorsilva:
Fixes https://github.com/concourse/concourse/issues/4586

We use the same task to build the alpine and ubuntu linux-rc's. They
both output the same filename and upload to the same gcs bucket
location. Ubuntu always wins because it's larger and takes longer to
upload. Last write = winner.

Screen Shot 2020-01-20 at 3 59 01 PM

This change ensures that the file names are different. Alpine based image
will continue to end with `-linux` and ubuntu one's should end in
`linux-ubuntu` now.

I feel confident that there are no further upstream mix-ups. I looked at the `build-rc-image` task and it still maps everything properly.

```yaml
- name: build-rc-image
  public: true
...
  - in_parallel:
      fail_fast: true
      steps:
        - task: build-alpine
          file: concourse-docker/ci/build-image.yml
          image: oci-build-task
          output_mapping: {image: image-alpine}
          privileged: true
        - task: build-ubuntu
          file: concourse-docker/ci/build-image.yml
          image: oci-build-task
          input_mapping: {linux-rc: linux-rc-ubuntu} #this PR ensures the task gets the correct rc now
          output_mapping: {image: image-ubuntu}
...
```

### DIFF
--- a/tasks/scripts/concourse-build
+++ b/tasks/scripts/concourse-build
@@ -7,7 +7,7 @@ export GOPATH=$PWD/gopath
 export PATH=$PWD/gopath/bin:$PATH
 
 if [ -z "${PLATFORM}" ]; then
-  echo "usage: export PLATFORM=<platform> $0" >&2
+  echo "usage: PLATFORM=<platform> $0" >&2
   exit 1
 fi
 
--- a/tasks/scripts/concourse-build
+++ b/tasks/scripts/concourse-build
@@ -6,15 +6,13 @@ set -e -x
 export GOPATH=$PWD/gopath
 export PATH=$PWD/gopath/bin:$PATH
 
-if [ "$#" -lt 1 ]; then
-  echo "usage: $0 <platform>" >&2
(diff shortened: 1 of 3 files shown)


tasks/scripts/concourse-build
```diff
@@ -6,15 +6,13 @@ set -e -x
 export GOPATH=$PWD/gopath
 export PATH=$PWD/gopath/bin:$PATH
 
-if [ "$#" -lt 1 ]; then
-  echo "usage: $0 <platform>" >&2
+if [ -z "${PLATFORM}" ]; then
+  echo "usage: export PLATFORM=<platform> $0" >&2
```
cirocosta:
```suggestion
  echo "usage: PLATFORM=<platform> $0" >&2
```

the shell will already take care of placing the environment variable for the executable being run

cirocosta:
goood catch, just left a little `nit` 😁  thanks!

PR #261, opened 2020-01-23
Update concourse & release pipelines

xtreme-sameer-vohra:
Currently our pipelines ( Concourse & Release ) block topgun tests from being run when the `bosh-check-props` job fails. 

This is not desirable as it prevents us from getting quick feedback from the topgun tests. 

The `bosh-check-props` will still prevent shipit from being run until it passes.

### DIFF
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -832,13 +832,11 @@ jobs:
   plan:
   - in_parallel:
     - get: concourse
-      passed: [bosh-check-props]
       trigger: true
     - get: unit-image
     - get: version
-      passed: [bosh-check-props]
     - get: linux-rc-ubuntu
-      passed: [build-rc, bosh-check-props]
+      passed: [build-rc]
       trigger: true
     - get: windows-rc
       passed: [build-rc]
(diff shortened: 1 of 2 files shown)


pivotal-jamie-klassen:
I totally see the desire for this in a certain scenario -
"I'm trying get my changes to the concourse binary tested in topgun, but some other jerk added some flags and forgot to add  them to the bosh packaging and now i can't run my tests!"
However, once that jerk adds the flags into the bosh packaging, they run the risk of breaking something else, and topgun will still need to re-run with their changes... I guess you still want to see topgun run (even if it's not running with the "final" version of the bosh release), but this begs the question - what flows is topgun testing that are so important you need the feedback even if you know you're using the wrong version of the bosh release? @pivotal-izabela-gomes @xtreme-sameer-vohra does this question make sense?

xtreme-sameer-vohra:
You're right, where if the packaging ( Helm or BOSH release ) is updated, we'd want to re-run with the changes to ensure there is no new breakage. 

We'd still desire to see topgun suite run as it allows us to continue validating the concourse component ( one of the components under test ) is devoid of issues. 

And maybe more relevant to this discussion, we're also stating that, the `check-props` jobs should be tested in parallel as opposed to serially with the topgun suite as the two are independent of each other.

haha love that `some jerk` reference ;p

wdyt ? @pivotal-jamie-klassen

PR #264, opened 2020-02-06
scripts: do not ignore USE_CONTAINERD flag

cirocosta:
with https://github.com/concourse/concourse/pull/5114, we're at a point
where we can start testing containerd-based workers out.

this commit then removes the flag that makes it all possible from the
set of flags - this way, we make sure we have it right at least in the chart.


cirocosta:
https://github.com/concourse/concourse-chart/pull/59 brings the flag to the chart

PR #269, opened 2020-02-10
scripts: add `osusergo` to tags when building fly

cirocosta:
with the introduction of the pure-go ZSTD lib (#5024), there might be
something new that's importing `os/user`, which ends up forcing a
dynamic build.

by adding the `osusergo` tag to the build of `fly`, we can instead
leverage the go-based implementation, and thus, have a static build of
`fly`.

### DIFF
--- a/tasks/scripts/fly-build
+++ b/tasks/scripts/fly-build
@@ -27,7 +27,7 @@ pushd concourse
     fi
   fi
 
-  go build --tags netgo -a -ldflags "$ldflags" -o $output/fly ./fly
+  go build --tags 'netgo osusergo' -a -ldflags "$ldflags" -o $output/fly ./fly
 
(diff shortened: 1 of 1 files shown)


tasks/scripts/fly-build
```diff
@@ -27,7 +27,7 @@ pushd concourse
     fi
   fi
 
-  go build --tags netgo -a -ldflags "$ldflags" -o $output/fly ./fly
+  go build --tags 'netgo osusergo' -a -ldflags "$ldflags" -o $output/fly ./fly
```
cirocosta: > When cgo is available, cgo-based (libc-backed) code is used by default. This can be overridden by using osusergo build tag, which enforces the pure Go implementation. 

(see https://golang.org/pkg/os/user/)

Issue #286, opened 2020-03-09
Dynamically determine postgres/helm version used in unit image


pivotal-izabela-gomes:
We want to be able to easily determine the versions that were used during test jobs for Postgresql and Helm. Currently, the process of determining those versions is manual, and it consists in looking at build logs for specific installed packages as the links below show. 

https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/unit-image/builds/10#L5e48eb1b:1711
https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/unit-image/builds/10#L5e48eb1b:514
https://ci.concourse-ci.org/teams/main/pipelines/release-5.5.x/jobs/k8s-topgun/builds/17.1

pivotal-izabela-gomes:
The job `discover-component-version` was mistakenly placed in `develop` group, instead of in the `publish` group.

pivotal-izabela-gomes:
Also, the jobs `discover-component-version` are currently in an error state in the release pipelines. The message says `ci/tasks/discover-component-version.yml` path cannot be found and, this is happening because the git resource `ci` is missing from the job.

PR #288, opened 2020-03-10
Add job to discover installed versions of Postgresql and Helm in the unit image

pivotal-izabela-gomes:
Added job `discover-components-version` to find out installed versions of Postgresql and Helm in the unit image.

- Created GCS bucket called `concourse-components-version`
- Added job that saves two files to the bucket. One for postgres
and the other for helm. Filename consists of [app name]-version-[concourse version].txt
and the content is the actual app version installed in the unit image
(eg. helm-version-5.5.8.txt has v2.16.3 as its contents).
- Used the same filename convention as files in the `concourse-ubuntu-dpkg-list` bucket

Solves issue https://github.com/concourse/ci/issues/286

Signed-off-by: Izabela Gomes <igomes@pivotal.io>

### DIFF
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -606,37 +606,6 @@ jobs:
             additional_tags: version/version
   on_failure: *failed-concourse
 
-- name: discover-component-version
-  public: true
-  serial: true
-  plan:
-  - in_parallel:
-    - get: concourse
-      passed: [build-rc-image]
-      trigger: true
-    - get: unit-image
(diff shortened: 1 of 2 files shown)


xtreme-sameer-vohra:
Hey @pivotal-izabela-gomes @pivotal-jamie-klassen 
This is great, nice to not have to do manual detective work :)

I am on the crypto CVE track and was wrapping up and thought this PR is useful our main pipeline as well. Added a PR #302 to address that.

Issue #284, opened 2020-03-06
Test linux binary with containerd enabled


muntac:
## What challenge are you facing?

containerd support for Concourse workers is being tested by running unit and integrations tests. The Guardian backend is also tested by running smoke tests against the packaged binary but the same is not done for containerd.

## What would make this better?

Have an equivalent of the `bin-smoke` job using containerd as the container backend for workers.This would involve bringing up the worker with the containerd flag, `CONCOURSE_GARDEN_USE_CONTAINERD`, set to true.

PR #289, opened 2020-03-11
Add bin-smoke-containerd 

xtremerui:
implements #284 

For smoke test concourse binary with containerd as worker backend

### DIFF
--- a/deployments/smoke/backend.tf
+++ b/deployments/smoke/backend.tf
@@ -1,7 +1,4 @@
-terraform {
-  backend "gcs" {
-    credentials = "keys/gcp.json"
-    bucket      = "concourse-smoke-state"
-    prefix      = "terraform/state"
-  }
+backend "gcs" {
+  credentials = "keys/gcp.json"
+  prefix      = "terraform/state"
 }
--- a/deployments/smoke/backend.tf
+++ b/deployments/smoke/backend.tf
@@ -1,4 +1,7 @@
-backend "gcs" {
-  credentials = "keys/gcp.json"
-  prefix      = "terraform/state"
+terraform {
+  backend "gcs" {
+    credentials = "keys/gcp.json"
+    bucket      = "concourse-smoke-state"
+    prefix      = "terraform/state"
+  }
 }
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -27,6 +27,7 @@ groups:
   - build-rc
   - build-rc-image
   - bin-smoke
+  - bin-smoke-containerd
   - upgrade
   - downgrade
 
@@ -783,6 +784,39 @@ jobs:
     image: unit-image
     file: ci/tasks/smoke.yml
     input_mapping: {endpoint-info: outputs}
+  on_success: *fixed-concourse
+  on_failure: *broke-concourse
+
+- name: bin-smoke-containerd
+  public: true
+  serial: true
+  plan:
+    - in_parallel:
+        - get: concourse
+          passed: [build-rc]
+          trigger: true
+        - get: version
+          passed: [build-rc]
+          trigger: true
+        - get: linux-rc-ubuntu
+          passed: [build-rc]
+          trigger: true
+        - get: unit-image
+        - get: ci
+    - task: terraform-smoke
+      image: unit-image
+      file: ci/tasks/terraform-smoke.yml
+      input_mapping: {linux-rc: linux-rc-ubuntu}
+      params:
+        GCP_PROJECT: cf-concourse-production
+        GCP_KEY: ((concourse_smoke_gcp_key))
+        SSH_KEY: ((concourse_smoke_ssh_key))
+        DEPLOYMENT: smoke
+        USE_CONTAINERD: true
+    - task: smoke
+      image: unit-image
+      file: ci/tasks/smoke.yml
+      input_mapping: {endpoint-info: outputs}
   on_success: *fixed-concourse
   on_failure: *broke-concourse
 
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -758,6 +758,7 @@ jobs:
 - name: bin-smoke
   public: true
   serial: true
+  serial_groups: [bin-smoke]
   plan:
   - in_parallel:
     - get: concourse
@@ -790,6 +791,7 @@ jobs:
 - name: bin-smoke-containerd
   public: true
   serial: true
+  serial_groups: [bin-smoke]
   plan:
     - in_parallel:
         - get: concourse
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -813,6 +813,7 @@ jobs:
         SSH_KEY: ((concourse_smoke_ssh_key))
         DEPLOYMENT: smoke
         WORKSPACE: containerd
+        TF_VAR_USE_CONTAINERD: "true"
     - task: smoke
       image: unit-image
       file: ci/tasks/smoke.yml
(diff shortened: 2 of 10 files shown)


tasks/scripts/concourse-build
```diff
@@ -43,7 +43,7 @@ mv concourse/concourse $bin
 [ -d runc ] && install -m 0755 runc/runc.amd64 $bin/runc
 [ -d cni ] && tar -zvxf cni/cni-plugins-linux-amd64-*.tgz -C $bin
 
-if [ "$PLATFORM" = "linux" ]; then
+if [[ $PLATFORM = *linux* ]]; then
```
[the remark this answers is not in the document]
xtremerui: Good point. Updated!

xtremerui:
If the goal of bin-smoke test doesn't include "ability to deploy on same env with different worker backend", then yeah dedicate VM for each kind makes sense. Another benefit of that would be these two jobs can now run in parallel. Will update the PR accordingly.

xtremerui:
PR updated @pivotal-jamie-klassen let me know wdyt about "-backend-config" part. Thx!

pipelines/concourse.yml
```diff
@@ -786,6 +787,39 @@ jobs:
...
+    - task: terraform-smoke
+      image: unit-image
+      file: ci/tasks/terraform-smoke.yml
+      input_mapping: {linux-rc: linux-rc-ubuntu}
+      params:
+        GCP_PROJECT: cf-concourse-production
+        GCP_KEY: ((concourse_smoke_gcp_key))
+        SSH_KEY: ((concourse_smoke_ssh_key))
+        DEPLOYMENT: smoke
+        USE_CONTAINERD: true
```
YoussB:
```suggestion
        USE_CONTAINERD: true
        WORKSPACE: containerd
```

tasks/scripts/terraform-smoke
```diff
@@ -18,15 +18,20 @@ echo "$SSH_KEY" > keys/id_rsa
 chmod 0600 keys/id_rsa
 ssh-keygen -y -f keys/id_rsa > keys/id_rsa.pub
 
-terraform init
+if [ "$USE_CONTAINERD" = true ]; then
+  terraform init -backend-config="bucket=concourse-smoke-state-containerd"
+else
+  terraform init -backend-config="bucket=concourse-smoke-state"
+fi
```
YoussB:
```suggestion
terraform init
```

deployments/smoke/backend.tf
```diff
@@ -1,7 +1,4 @@
-terraform {
-  backend "gcs" {
-    credentials = "keys/gcp.json"
-    bucket      = "concourse-smoke-state"
-    prefix      = "terraform/state"
-  }
+backend "gcs" {
+  credentials = "keys/gcp.json"
+  prefix      = "terraform/state"
```
YoussB:
```suggestion
terraform {
  backend "gcs" {
    credentials = "keys/gcp.json"
    bucket      = "concourse-smoke-state"
    prefix      = "terraform/state"
  }
```

pipelines/concourse.yml
xtremerui: where does `containerd` got consumed? But I got your point, that is cleaner than using -backend-config. I will try that thx!

xtremerui: NVM. I saw the script is already consuming the env var. Neat!

deployments/smoke/backend.tf
```diff
@@ -1,7 +1,5 @@
-terraform {
```
pivotal-jamie-klassen: I think this introduces a syntax error?

tasks/scripts/terraform-smoke
```diff
@@ -18,8 +18,6 @@ echo "$SSH_KEY" > keys/id_rsa
 chmod 0600 keys/id_rsa
 ssh-keygen -y -f keys/id_rsa > keys/id_rsa.pub
 
-terraform init
```
pivotal-jamie-klassen: I think this line is still necessary since it synchronizes with the `.tfstate` file in the backend.

deployments/smoke/backend.tf
```diff
@@ -1,7 +1,5 @@
...
-  backend "gcs" {
-    credentials = "keys/gcp.json"
-    bucket      = "concourse-smoke-state"
-    prefix      = "terraform/state"
-  }
+backend "gcs" {
+  credentials = "keys/gcp.json"
+  bucket      = "concourse-smoke-state"
+  prefix      = "terraform/state"
```
pivotal-jamie-klassen: Accordingly, you probably don't need to change the indentation here

xtremerui:
@pivotal-jamie-klassen we could also utilized systemd service template https://fedoramagazine.org/systemd-template-unit-files/

pivotal-jamie-klassen:
@xtremerui it looks like that might be strictly interpolation... in our case I think we still need a conditional.

xtremerui:
Could the conditianal logic live here https://github.com/concourse/ci/blob/33691195c11c5b5eacb311701a390d97b6179d43/deployments/smoke/smoke.tf#L189-L192

taylorsilva:
@jomsie and I re-implemented the `use_containerd` var and threaded it down to the systemd service template. We did not commit it on this branch, leaving the diff here for feedback. We didn't get to test it either, needed to get a gcp json key to run any terraform commands.

```patch
diff --git a/deployments/smoke/smoke.tf b/deployments/smoke/smoke.tf
index cbef3dd..5a6000d 100644
--- a/deployments/smoke/smoke.tf
+++ b/deployments/smoke/smoke.tf
@@ -17,6 +17,11 @@ variable "use_https" {
   default = true
 }

+variable "USE_CONTAINERD" {
+  type    = bool
+  default = false
+}
+
 provider "google" {
   credentials = "keys/gcp.json"
   project     = var.project
@@ -125,7 +130,7 @@ data "template_file" "worker_conf" {
   template = file("systemd/smoke-worker.conf.tpl")

   vars = {
-    workspace = terraform.workspace
+    use_containerd = var.USE_CONTAINERD
   }
 }

diff --git a/deployments/smoke/systemd/smoke-worker.conf.tpl b/deployments/smoke/systemd/smoke-worker.conf.tpl
index 14a4f6f..f73652a 100644
--- a/deployments/smoke/systemd/smoke-worker.conf.tpl
+++ b/deployments/smoke/systemd/smoke-worker.conf.tpl
@@ -2,7 +2,7 @@
 Environment=CONCOURSE_WORK_DIR=/etc/concourse/work-dir
 Environment=CONCOURSE_TSA_PUBLIC_KEY=/etc/concourse/host_key.pub
 Environment=CONCOURSE_TSA_WORKER_PRIVATE_KEY=/etc/concourse/worker_key
-%{ if workspace == "containerd" ~}
+%{ if use_containerd == "true" ~}
 Environment=CONCOURSE_GARDEN_USE_CONTAINERD="true"
 %{ else ~}
 Environment=CONCOURSE_GARDEN_CONFIG=/etc/concourse/garden.ini
diff --git a/pipelines/concourse.yml b/pipelines/concourse.yml
index 78b7666..1a47559 100644
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -813,6 +813,7 @@ jobs:
         SSH_KEY: ((concourse_smoke_ssh_key))
         DEPLOYMENT: smoke
         WORKSPACE: containerd
+        TF_VAR_USE_CONTAINERD: "true"
     - task: smoke
       image: unit-image
       file: ci/tasks/smoke.yml
diff --git a/tasks/terraform-smoke.yml b/tasks/terraform-smoke.yml
index ef7a899..94be046 100644
--- a/tasks/terraform-smoke.yml
+++ b/tasks/terraform-smoke.yml
@@ -11,6 +11,7 @@ params:
   SSH_KEY:
   WORKSPACE: default
   USE_HTTPS: 'true'
+  TF_VAR_USE_CONTAINERD: "false"

 inputs:
 - name: concourse
```

taylorsilva:
@xtremerui and I applied the above patch.

PR #310, opened 2020-04-27
automate uploading bosh release to tanzu network

pivotal-jamie-klassen:
I have successfully tested this on pivnet-integration.cfapps.io by modifying
the pipeline template slightly and changing release-5.5.x so that the latest
version that has passed topgun gets uploaded.

https://pivnet-integration.cfapps.io/products/p-concourse#/releases/633803

In order to see this at work we'd need to wait until the next time `shipit`
gets used, since part of this change involves threading the stemcell.

looking forward, it would be nice to parameterize/dynamically determine:
* the release notes URL
* the stemcell family
* the supported versions of Postgres

### DIFF
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -977,6 +977,8 @@ jobs:
       passed: [publish-bosh-release]
   - task: create-pivnet-metadata
     file: ci/tasks/create-pivnet-metadata.yml
+    params:
+      END_OF_GENERAL_SUPPORT: ((end_of_general_support))
   - in_parallel:
     - put: pivnet-bosh-release
       inputs:
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -5,6 +5,7 @@
 #                                       concourse matches the desired release version
 #   ((concourse_smoke_deployment_name)) a unique name for the smoke bosh deployment
 #   ((latest_release))                  the latest concourse/concourse tag for upgrade testing (choose this based on the patch release eg 5.5 for 5.5.x)
+#   ((end_of_general_support))          YYYY-MM-dd format when support for this minor line ends
 #
 # the following git branches need to be created:
 #
@@ -29,6 +30,12 @@ resource_types:
 - name: slack-notifier
   type: registry-image
   source: {repository: mockersf/concourse-slack-notifier}
+
+- name: pivnet
+  type: registry-image
+  source:
+    repository: pivotalcf/pivnet-resource
+    tag: latest-final
 
 groups:
 - name: develop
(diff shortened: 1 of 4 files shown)


pipelines/release.yml
```diff
@@ -1341,6 +1375,15 @@ resources:
...
     regexp: ((release_minor)).*
 
+- name: pivnet-bosh-release
+  type: pivnet
+  icon: chart-timeline-variant
+  source:
+    api_token: ((pivnet-token))
+    product_slug: p-concourse
+    copy_metadata: true
+    endpoint: https://network.pivotal.io
```
aoldershaw, line 1385: Not sure if you know, but I imagine that this endpoint will change at some point to not be under `pivotal.io` - might it make sense to not define the `endpoint`? The default is already `https://network.pivotal.io`, and I speculate they'll update that if Tanzu network gets moved elsewhere

aoldershaw: Though `https://network.pivotal.io` may stick around for a while, I'm not sure

pivotal-jamie-klassen: nice, I didn't realize it had a default!

aoldershaw:
Another question: with this change, we'll be uploading 5.7.x and 5.8.x to Pivnet as well. Granted, it's available for Admins only, but is this something we want to be doing?

tasks/scripts/create-pivnet-metadata
```diff
@@ -0,0 +1,21 @@
...
+  eula_slug: pivotal_software_eula
+  description: |
+    Concourse for VMware Tanzu v$(cat version/version) is a stable and VMware supported version of Concourse. Please refer to Release Notes for breaking changes, features and fixes in this release.
+    
+    Concourse for VMware Tanzu v$(cat version/version) was tested on Stemcell $(cat gcp-xenial-stemcell/version) (Xenial) upon release and supports the 621.* Stemcell family
+    
+    Concourse for VMware Tanzu v$(cat version/version) requires PostgreSQL 9.5+ 
+  release_notes_url: https://docs.pivotal.io/p-concourse/v5/rn/
+  availability: Admins Only
+  end_of_support_date: ((end_of_general_support))
```
aoldershaw, line 17: How does this variable interpolation work, out of curiosity?

pivotal-jamie-klassen: hahaha it doesn't! this used to be inline and I blindly copy-pasted it into an external script

aoldershaw: Okay phew! Thought there was some witchcraft going on somewhere!

pivotal-jamie-klassen: "I have successfully tested this". yep, right before i changed it a bunchh.

pivotal-jamie-klassen: fixed 56630e45c5da0eae86c04a245deb125dfffb7969

PR #312, opened 2020-04-30
Fix path to file containing status of Norsk pipelines' jobs

pivotal-izabela-gomes:
Currently, the script that reads status.json from [oslo-scan-status](https://github.com/pivotal/oslo-scan-status) repo doesn't take into consideration the different folders that exist based on pipeline names for concourse. So, the 3 existing jobs that should monitor norsk pipelines for different Concourse versions (p-concourse, p-concourse-4.2.x and p-concourse-5.5.x) read the file under p-concourse folder instead of their respective folder (named after each pipeline). The git resource properly watches for changes in the correct path, but the status.json file that the task reads from is actually the same for all 3 jobs. The PR fixes that issue by passing a param with the pipeline name to the task/script. 

The PR also changes the format of slack messages (for norsk-mirror related jobs) to not use the `yarn blame` method, since we read the status.json from a [repo](https://github.com/pivotal/oslo-scan-status) that gets updated by a bot. It now sends a message to slack with urls of builds that fail in our norsk pipelines.

### DIFF
--- a/pipelines/norsk-mirror.yml
+++ b/pipelines/norsk-mirror.yml
@@ -46,6 +46,8 @@ jobs:
   - get: ci
   - task: check-status-file
     file: ci/tasks/check-norsk-status.yml
+    params:
+      NORSK_PIPELINE: p-concourse
     input_mapping: {scan-status: latest-scan-status}
   on_success:
     do:
@@ -81,6 +83,8 @@ jobs:
   - get: ci
   - task: check-status-file
     file: ci/tasks/check-norsk-status.yml
+    params:
+      NORSK_PIPELINE: p-concourse-4.2.x
     input_mapping: {scan-status: 4.2.x-scan-status}
   on_success:
     do:
@@ -116,6 +120,8 @@ jobs:
     - get: ci
     - task: check-status-file
       file: ci/tasks/check-norsk-status.yml
+      params:
+        NORSK_PIPELINE: p-concourse-5.5.x
       input_mapping: {scan-status: 5.5.x-scan-status}
   on_success:
     do:
--- a/pipelines/norsk-mirror.yml
+++ b/pipelines/norsk-mirror.yml
@@ -51,25 +51,17 @@ jobs:
     input_mapping: {scan-status: latest-scan-status}
   on_success:
     do:
-    - task: format-slack-message
-      file: ci/tasks/format-slack-message.yml
-      input_mapping: {src: latest-scan-status}
-      params:
-        GITHUB_TOKEN: ((concourse_github_dummy.access_token))
-        SLACK_TOKEN: ((slack_token))
     - put: notify
       params:
-        message_file: message/message
         mode: normal
         alert_type: fixed
   on_failure:
     do:
-    - task: format-slack-message
-      file: ci/tasks/format-slack-message.yml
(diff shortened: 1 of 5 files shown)


tasks/scripts/format-slack-message-norsk
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e -x
+
+message_file="$PWD/message/message"
+echo $(jq -r '.jobs[] | select(.status == "failed") | .url' $PWD/src/$NORSK_PIPELINE/status.json) > $message_file
```
YoussB, line 6: unless the `echo` is intentional, I think we can get rid of it since in case of multiple failures, this will remove the newlines between each line.
```suggestion
jq -r '.jobs[] | select(.status == "failed") | .url' $PWD/src/$NORSK_PIPELINE/status.json > $message_file
```

pipelines/norsk-mirror.yml
```diff
@@ -81,28 +75,22 @@ jobs:
   - get: ci
   - task: check-status-file
     file: ci/tasks/check-norsk-status.yml
+    params:
+      NORSK_PIPELINE: p-concourse-4.2.x
```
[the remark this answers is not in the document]
pivotal-izabela-gomes: Probably not... I left the job there just because the norsk pipeline for 4.2.x still exists. But I'll check with Scott if it's really necessary to keep an eye on it or not.

tasks/scripts/format-slack-message-norsk
pivotal-izabela-gomes: The echo was intentional, and the idea behind it was to list the urls to the jobs on the norsk pipeline that failed. The default slack message would lead us to our norsk-mirror job (which only contains a task that scans the status.json file for failed jobs). I wanted to make it easy to jump to the pipeline/job that actually failed by clicking on the slack message. But I guess you have a good point... if all jobs failed, we would have a slack message with probably more than 10 urls. 
Instead of removing the echo, I thought that I could include then only the url to the nork pipeline, not all those jobs that failed. What do you think? If you agree I'll submit changes to the PR later today or tomorrow.

YoussB: aaand it took me that long to respond 😁.
my idea was that the urls wouldn't have any separations between them so if we have multiple links it will be a bit unreadable.

YoussB: but yeah, maybe using another link could solve it as well.

pipelines/norsk-mirror.yml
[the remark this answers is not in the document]
YoussB: I think we don't need to remove it using this PR anyway, I was just wondering while we're at it. maybe we can ask OSL to remove the pipeline as well it is not needed.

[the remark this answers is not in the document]
pivotal-izabela-gomes: I'll check with OSL if they can remove it or if they keep for historical reasons.

tasks/scripts/format-slack-message-norsk
pivotal-izabela-gomes: Good catch! Thanks!

PR #315, opened 2020-05-07
Ensure reconfigure pipeline configures itself first

taylorsilva:
This PR adds a job that reconfigures the reconfigure pipeline itself. It avoids the situation where a new pipeline is added but it isn't actually added until the build is re-run.

Example: https://ci.concourse-ci.org/teams/main/pipelines/reconfigure-pipelines/jobs/reconfigure-pipelines/builds/164.1
image

This new first job still has this problem but at least it's isolated now
and won't occur for the majority of use-cases. We're not likely to change the first job often/ever.

### DIFF
--- a/pipelines/reconfigure.yml
+++ b/pipelines/reconfigure.yml
@@ -43,10 +43,18 @@ resources:
     uri: https://github.com/concourse/oxygen-mask
 
 jobs:
+- name: reconfigure-self
+  plan:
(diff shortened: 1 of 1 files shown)


pivotal-jamie-klassen:
Welp, I sure meant to get around to this ages ago 😬 https://github.com/orgs/concourse/projects/36#card-25085021

PR #320, opened 2020-05-12
Move PRs pipeline from main team to contributor team

xtremerui:
See context on https://github.com/concourse/hush-house/pull/127

For security contributors should be limited to operator PRs pipeline only. With current implementation of Concourse RBAC configuration it can't do that if PRs pipeline stays in the
same team with others.

So this pull request moves PRs pipeline from main team to contributor team. Along with it I have to extract the reconfigure task for PRs pipeline from reconfigure.yml to its own since currently set_pipeline step will only set pipeline to current team.

For dev ops when setting reconfigure-contributor.yml, `fly login -n contributor` should be done first.

### DIFF
--- a/pipelines/reconfigure-contributor.yml
+++ b/pipelines/reconfigure-contributor.yml
@@ -1,0 +1,31 @@
+resource_types:
+- name: concourse-pipeline
+  type: registry-image
+  source:
+    repository: concourse/concourse-pipeline-resource
+
+resources:
+- name: pipelines-and-tasks
+  type: git
+  icon: github-circle
+  source:
+    uri: https://github.com/concourse/ci
+    paths:
+    - pipelines
+    - tasks
+
+jobs:
+- name: reconfigure-self
+  plan:
+  - get: pipelines-and-tasks
+    trigger: true
+  - set_pipeline: reconfigure-pipelines
+    file: pipelines-and-tasks/pipelines/reconfigure-contributor.yml
+
+- name: reconfigure-pipelines
+  plan:
+  - get: pipelines-and-tasks
+    trigger: true
+    passed: [reconfigure-self]
+  - set_pipeline: prs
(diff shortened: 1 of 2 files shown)


pipelines/reconfigure-contributor.yml
```diff
@@ -0,0 +1,31 @@
+resource_types:
+- name: concourse-pipeline
+  type: registry-image
+  source:
+    repository: concourse/concourse-pipeline-resource
+
```
pivotal-jamie-klassen, line 6: I imagine you copy-pasted, but I can't help but noticed this is unused!
```suggestion
```

cirocosta:
> For dev ops when setting reconfigure-contributor.yml, fly login -n contributor should be done first.

Initially, when reading this snippet, I thought "hmm, that sounds like something that should be documented in boarding-pass", but now I wonder if this is something that we should capture as part of the automation that we're working on for `greenpeace` :thinking:

xtremerui:
@cirocosta it should be configured whenever we set reconfigure pipeline in main team, in greenpease I'd imagine.

pipelines/reconfigure-contributor.yml
xtremerui: Interesting I wonder why the pipeline validation didn't capture this unused resource type

cirocosta:
> @cirocosta it should be configured whenever we set reconfigure pipeline in main team, in greenpeace I'd imagine.

@aoldershaw, do you think we should have an issue/note for that? not sure if that's something we do already :grin:

aoldershaw:
@cirocosta we're actually trying to figure out the best approach for bootstrapping CI's pipelines right now. I'll create an issue for this

xtremerui:
@aoldershaw for the record another thing I did was copy `concourse/main/pull_requests_access_token` to `concourse/contributor/pull_requests_access_token` under vault.

PR #340, opened 2020-06-05
Mark old issues as stale and close shortly after

muntac:
Issues will be labelled `stale` after no activity for 6 weeks, and then be closed a week after that.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,19 @@
+name: Mark stale issues
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v3
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
...
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
```
jamieklassen, line 13: I see there's a [v2](https://github.com/actions/stale/tree/v2) and [v3](https://github.com/actions/stale/tree/v3) of this action -- any reason for choosing v1? I think v3 might be pretty cool because it removes the stale label on issues when there is recent activity: https://github.com/actions/stale/releases/tag/v3.0.0.

muntac: Good catch! Fixed.

PR #341, opened 2020-06-12
Add pipeline to detect changes in Datadog dashboards

pivotal-izabela-gomes:
I created this pipeline so it detects changes in our Datadog dashboards, and once we are notified we can take action to describe those changes as code in our [Terraform module](url). 

There is one job per Concourse deployment (CI and Hush House). One task uses `terraform import` to retrieve the existing resources/dashboards from Datadog (previously created by the tf module in greenpeace repo) and save into a terraform.tfstate file. The subsequent task takes in the tf state file and runs `terraform plan` to find differences between dashboards in Datadog and our code (terraform module) that originally created those dashboards. If changes are detected, we want to notify the team via slack so we can update our Terraform module with recent changes.


pivotal-izabela-gomes:
I'm merging this PR because it shouldn't affect existing pipelines.

PR #349, opened 2020-07-13
Switch to ytt for pipeline templating

taylorsilva:
This PR switches us from using jsonnet to ytt for pipeline templating. 

I verified that the rendered pipeline configs match the current
pipelines in ci.concourse-ci.org by doing `fly set-pipeline` on each
pipeline and checking that no diff appeared.


YoussB:
whoops! looks like it took me a long time to get to this that it got merged 😁. I will still take a look.

pipelines/resources/template.yml
```diff
@@ -0,0 +1,427 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:template", "template")
+#@ load("@ytt:assert", "assert")
+
+#@ resource_name = data.values.resource_name or assert.fail("resource name must be specified")
+
+#@ def build_params(resource_name):
+  #@ if resource_name == "registry-image":
```
YoussB, line 8: would it make sense to have the if/elses only for the build args?

taylorsilva: almost! Looked it over and almost said yes but the last `elseif` is this:
```
  #@ elif resource_name == "concourse-pipeline":
build: "."
```
which is the one different case. Looks like we'd have to change the dockerfile for that resource: https://github.com/concourse/concourse-pipeline-resource#using-a-dockerfile

or we could work on removing it completely from Concourse since we have set_pipeline now 😄

PR #352, opened 2020-07-15
ignore environment variables for experimental features

muntac:
When a feature is its experimental phase (e.g. containerd) putting the corresponding environment variables in published BOSH releases and Helm charts can potentially confuse users. This is because the feature can be removed or changed (e.g. environment variable name changes) before it is becomes ready for GA. At which point the chart or release has to be updated again with a name change. The flags should instead be included as part of making the feature GA.


muntac:
Closing this in favor of https://github.com/concourse/ci/pull/353.

PR #355, opened 2020-07-30
Integrate release note generator

clarafu:
A new `create-draft-release` job was added into the publish group of the release pipeline. This job is responsible for creating a draft release in the Concourse repository with a generated release note.

The release note is generated using the `release-me` CLI https://github.com/clarafu/release-me. It generates a release note using all the PRs merged after the last release of the current release branch and before the commit that has made it through the testing part of the pipeline. It can be used to generate a release note for release branches, where the last release it will use will be the latest release made on that release branch. For example, if I had a release/5.5.x branch and the latest release is 5.5.1, the release note generated will include all prs merged after the 5.5.1 release and before the last commit that has been successfully tested. 

The generated release note will include all the PR titles and optional descriptions and be split into multiple sections. It is possible for this job to fail if one of the prs that got merged in does not have the appropriate labels (which are used to split it into sections). 

This `create-draft-release` job will run for every commit that makes it through the testing part of the release pipeline and will create a new draft release for every commit (which is more like it will update the description for the draft release with the new pr). This will allow for the release anchor to then take a look at the generated draft release note and manually edit the release note within github. After they are happy with the release note, they can click the `ship-it` job which will then convert the draft release into a real release.

### DIFF
--- a/tasks/scripts/build-release-notes
+++ b/tasks/scripts/build-release-notes
@@ -1,4 +1,12 @@
 #!/bin/bash
 
+LAST_COMMIT_SHA = $(cat concourse/.git/ref)
+
 chmod +x release-me/releaseme
-./release-me/releaseme generate --github-token=$GITHUB_TOKEN --github-owner=concourse --github-repo=concourse > built-notes/notes.md
+./release-me/releaseme generate \
+  --github-token=$GITHUB_TOKEN \
+  --github-owner=concourse \
+  --github-repo=concourse \
+  --github-branch=$GITHUB_BRANCH \
+  --last-commit-SHA=$LAST_COMMIT_SHA \
+  > built-notes/notes.md
--- a/tasks/scripts/build-release-notes
+++ b/tasks/scripts/build-release-notes
@@ -1,14 +1,4 @@
 #!/bin/bash
 
-if [ -d version ]; then
-  version=v$(cat version/version)
-elif [ "$#" -eq 1 ]; then
-  version=$1
-else
-  echo "version must be specified, either as an argument or an input" >&2
-  exit 1
-fi
-
-echo "$version" > ./built-notes/release-name
-# if this fails, it's because we forgot to manually promote latest.md to $version.md
-cp release-notes/release-notes/"$version".md ./built-notes/notes.md
+chmod +x release-me/releaseme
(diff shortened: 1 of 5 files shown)


tasks/scripts/build-release-notes
```diff
@@ -1,14 +1,12 @@
...
-echo "$version" > ./built-notes/release-name
-# if this fails, it's because we forgot to manually promote latest.md to $version.md
-cp release-notes/release-notes/"$version".md ./built-notes/notes.md
+chmod +x release-me/releaseme
+./release-me/releaseme generate \
+  --github-token=$GITHUB_TOKEN \
+  --github-owner=concourse \
+  --github-repo=concourse \
+  --github-branch=$GITHUB_BRANCH \
+  --last-commit-SHA=$LAST_COMMIT_SHA \
```
jamieklassen, line 11: When I run `go run main.go --last-commit-SHA=foo` in the root of `clarafu/release-me`, I get:

```
Error: unknown flag: --last-commit-SHA
```

In 674bd6bd72514cb924714abda8844dd8e67eaa69 you mention this flag as a feature of release-me... maybe you just haven't pushed it on that repo yet?

clarafu: Oh sorry I haven't pushed to the repo yet! I just built the binary with the feature and shipped it

clarafu: I just pushed it to the repo!

clarafu:
@jamieklassen Unfortunately it will overwrite any release note changes made manually on the draft release if a new commit comes in. I currently don't have a good way of getting around that, the way that I was imagining the flow was that only when you are sure that this is the cut off for this release, you will edit the draft release note manually in github and then click ship-it. But I understand how things don't always go as well as planned, so maybe this is a too naive way of making the release process?

jamieklassen:
Does it feel too clunky to create the draft release on-demand as well? So that, before you can ship, you do:

1. run the `create-draft-release` job
1. make any manual edits to the draft release
1. run the `shipit` job

I guess this would require removing the `trigger: true` from the `get: concourse` step in the `create-draft-release` job.

clarafu:
I don't think it feels too clunky! I was deciding between the two options of having it automatically or manually create a draft release and we just decided on auto for the sake of possibly using it generate nightly releases, but I agree that with the current flow of allowing the release anchor to manually edit the draft release feels better if they have to manually run the `create-draft-release` job prior to it. I will change that!

PR #361, opened 2020-08-07
update terraform runtime selection logic

muntac:
~~CONCOURSE_USE_CONTAINERD has been replaced by CONCOURSE_RUNTIME. This unfortunately is going to break the `bin-smoke-containerd` test for 6.x pipelines from 6.4 and below. containerd is experimental in those releases and the job will be paused until we figure out a nice way to support the old 6.x pipelines and the current ones.~~

Update: The **non-breaking** way would be to pass the version number (`release_major`, `release_minor`) into the [release-v6.yml](https://github.com/concourse/ci/blob/3b5edb4a71908ea1c68aeb61f46f77385d4d98bb/pipelines/release-v6.yml#L704), and use that to select `CONCOURSE_USE_CONTAINERD` or `CONCOURSE_RUNTIME` accordingly.

I suggest doing that instead of the approach in this PR.

### DIFF
--- a/deployments/smoke/smoke.tf
+++ b/deployments/smoke/smoke.tf
@@ -18,13 +18,18 @@ variable "use_https" {
 }
 
 variable "image" {
-  type = string
+  type    = string
   default = "ubuntu-1804-bionic-v20181003"
 }
 
-variable "USE_CONTAINERD" {
+variable "ENABLE_CONTAINERD" {
   type    = bool
   default = false
+}
+
+variable "HAS_RUNTIME_FLAG" {
+  type    = bool
+  default = true
 }
 
 provider "google" {
@@ -135,7 +140,8 @@ data "template_file" "worker_conf" {
   template = file("systemd/smoke-worker.conf.tpl")
 
   vars = {
-    use_containerd = var.USE_CONTAINERD
+    enable_containerd = var.ENABLE_CONTAINERD
+    has_runtime_flag  = var.HAS_RUNTIME_FLAG
   }
 }
 
--- a/pipelines/reconfigure.yml
+++ b/pipelines/reconfigure.yml
@@ -280,16 +280,19 @@ jobs:
     vars:
       branch_name: algorithm-v3
       environment: algorithm
+      compatibility_version: 6.0.0
   - set_pipeline: runtime-drills
     file: pipelines-and-tasks/pipelines/branch.yml
     vars:
       branch_name: master
       environment: runtime-drills
+      compatibility_version: dev
   - set_pipeline: set-zstd-drills-env
     file: pipelines-and-tasks/pipelines/branch.yml
     vars:
       branch_name: use_zstd
       environment: use-zstd
+      compatibility_version: 6.0.0
   - set_pipeline: enterprise-helm-chart
     file: concourse-for-k8s/ci/pipeline.yml
     vars:
--- a/pipelines/release-v6.yml
+++ b/pipelines/release-v6.yml
@@ -671,6 +671,8 @@ jobs:
       trigger: true
     - get: unit-image
     - get: ci
+  - load_var: concourse_version
+    file: version/version
   - task: terraform-smoke
     image: unit-image
     file: ci/tasks/terraform-smoke.yml
@@ -681,6 +683,7 @@ jobs:
       SSH_KEY: ((concourse_smoke_ssh_key))
       WORKSPACE: release-((release_minor))
       USE_HTTPS: ((bin_smoke_use_https))
+      CONCOURSE_VERSION: ((.:concourse_version))
   - task: smoke
     image: unit-image
     file: ci/tasks/smoke.yml
@@ -693,17 +696,19 @@ jobs:
   serial: true
   plan:
     - in_parallel:
-        - get: concourse
-          passed: [build-rc]
-          trigger: true
-        - get: version
-          passed: [build-rc]
-          trigger: true
-        - get: linux-rc-ubuntu
-          passed: [build-rc]
-          trigger: true
-        - get: unit-image
-        - get: ci
+      - get: concourse
+        passed: [build-rc]
+        trigger: true
+      - get: version
+        passed: [build-rc]
+        trigger: true
+      - get: linux-rc-ubuntu
+        passed: [build-rc]
+        trigger: true
+      - get: unit-image
+      - get: ci
+    - load_var: concourse_version
+      file: version/version
     - task: terraform-smoke
       image: unit-image
       file: ci/tasks/terraform-smoke.yml
@@ -714,7 +719,8 @@ jobs:
         SSH_KEY: ((concourse_smoke_ssh_key))
         DEPLOYMENT: smoke
         WORKSPACE: release-((release_minor))-containerd
-        TF_VAR_USE_CONTAINERD: "true"
+        TF_VAR_ENABLE_CONTAINERD: "true"
+        CONCOURSE_VERSION: ((.:concourse_version))
     - task: smoke
       image: unit-image
       file: ci/tasks/smoke.yml
--- a/tasks/scripts/terraform-smoke
+++ b/tasks/scripts/terraform-smoke
@@ -10,6 +10,29 @@ if [ -d linux-rc ]; then
   cp linux-rc/concourse-*.tgz $deployment_path/concourse.tgz
 fi
 
+# takes 2 semver of form x.y.z[-rc.123] and returns
+# 0: =
+# 1: <
+# 2: >
+cmp() {
+  # strip pre-release (-rc) or build (+123) from semver
+  v1=$(echo $1 | sed -e 's/[-+].*$//g')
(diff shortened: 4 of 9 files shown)


chenbh:
Gonna take over this PR as @muntac is on vacation

jamieklassen:
@chenbh reading your commit message, I think the rename is pending for 6.5.0 rather than 6.4.2 -- does that sound right?

chenbh:
@jamieklassen that is absolutely correct

deployments/smoke/smoke.tf
```diff
@@ -220,3 +225,18 @@ output "guest_password" {
...
+  left = regex("(\\d)\\.(\\d)\\.(\\d).*", var.concourse_version)
+  x1   = local.left[0]
+  y1   = local.left[1]
+  z1   = local.left[2]
+
+  x2 = 6
+  y2 = 5
+  z2 = 0
+
+  gt_650 = (local.x1 > local.x2) || (local.x1 == local.x2 && local.y1 > local.y2) || (local.x1 == local.x2 && local.y1 == local.y2 && local.z1 > local.z2)
```
aoldershaw, line 241: This should be `gte`, right? (as should the corresponding conditions, not just the name)

jomsie: yeah from the (crossed out) pr and commit message(s) looks like it should be 6.5 and above (gte) using CONCOURSE_RUNTIME

jomsie:
I was just doing some testing with @chenbh's help, and I'm getting some issues when the minor gets into double digits.

```
❯ terraform refresh --var current=6.10.1-rc.2 --var target=6.10.1

Error: Error in function call

  on main.tf line 32, in locals:
  32:   left  = regex("(\\d)\\.(\\d)\\.(\\d).*", var.current)
    |----------------
    | var.current is "6.10.1-rc.2"

Call to function "regex" failed: pattern did not match any part of the given
string.


Error: Error in function call

  on main.tf line 36, in locals:
  36:   right = regex("(\\d).(\\d).(\\d)", var.target)
    |----------------
    | var.target is "6.10.1"

Call to function "regex" failed: pattern did not match any part of the given
string.
```


And also @chenbh it looks like it's treating 6.4.10 as 6.4.1

```
❯ terraform refresh --var current=6.4.10 --var target=6.4.1
data.template_file.worker_conf: Refreshing state... [id=eabbc95d17c4fc3fd44b4b712ac7425e71ada510622b690e1a862034ae7b10e3]

Outputs:

current = 6.4.1
file = [Service]
Environment=CONCOURSE_WORK_DIR=/etc/concourse/work-dir
Environment=CONCOURSE_TSA_PUBLIC_KEY=/etc/concourse/host_key.pub
Environment=CONCOURSE_TSA_WORKER_PRIVATE_KEY=/etc/concourse/worker_key
Environment=CONCOURSE_RUNTIME=containerd

gt = false
gteq = true
lt = false
lteq = true
target = 6.4.1
```

chenbh:
@jomsie good catch!

tasks/smoke.yml
```diff
@@ -9,6 +9,7 @@ inputs:
 - name: concourse
 - name: ci
 - name: endpoint-info
+- name: version
```
jamieklassen, line 12: `smoke.yml` gets consumed in many places:
* `bin-smoke`, `bin-smoke-kernel`, `bin-smoke-containerd`, `bosh-smoke` in `concourse.yml` - this pipeline is quite intentional about _not_ having a semver resource as it purely produces dev builds. How we gonna do a version comparison there?
* `bin-smoke`, `bin-smoke-containerd`, `bosh-smoke` in `release-v6.yml` - I think these all have a semver resource called `version` in their build context so we're probably safe here.
* various and sundry spots in branch/drills/legacy pipelines

jamieklassen:
In any case, I'm pausing https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/bin-smoke-containerd to reduce the notification fatigue. Once this PR gets resolved, please unpause!

tasks/scripts/terraform-smoke
```diff
@@ -26,8 +28,9 @@ terraform workspace select "$WORKSPACE" || \
 terraform apply \
   --auto-approve \
   --var project=$GCP_PROJECT \
-  --var use_https=$USE_HTTPS \
+  --var use_https=$USE_HTTPS
   --var image=$IMAGE
+  --var concourse_version=$CONCOURSE_VERSION
```
aoldershaw, line 33: I think you will need the linebreaks here:

```suggestion
  --var use_https=$USE_HTTPS \
  --var image=$IMAGE \
  --var concourse_version=$CONCOURSE_VERSION

```

tasks/smoke.yml
chenbh: Whoops, this change should've been in `terraform-smoke.yml` instead.

What do you think about making terraform dumb and only capable of emitting/omitting stuff based on flags. This way we can pass the flags directly to the task similar to how we  https://github.com/concourse/ci/blob/f8e639f72da8961327c83665843ef1cd03e226fa/pipelines/concourse.yml#L753

This would also allow us to move the version comparison logic into the bash script (which can set/unset env vars as needed)

- `concourse.yml`: uses default
- `release.yml` (I believe this is legacy release pipeline): add a `load_var` step to extract the version from semver resource and pass it in
- `release-v6.yml`: same as above
- `branch.yml`:  add a pipeline variable (`compatibility_version` or something) that has to be passed in
- `drill.yml`: not used by the reconfigure pipeline, but can use the same behaviour as above

pipelines/reconfigure.yml
```diff
@@ -296,16 +296,19 @@ jobs:
...
     vars:
       branch_name: master
       environment: runtime-drills
+      compatibility_version: dev
   - set_pipeline: set-zstd-drills-env
     file: pipelines-and-tasks/pipelines/branch.yml
     vars:
       branch_name: use_zstd
       environment: use-zstd
+      compatibility_version: 6.0.0
```
chenbh, line 311: I don't know if these numbers are accurate, they seemed to all be paused on CI and were never run at all. 

I'm assuming `algorithms-v3` is v6, `runtime-drills` seems to have replaced `drills.yml` (not sure why the file is still in this repo) and I have no idea what `use-zstd` is

pipelines/release-v6.yml
```diff
@@ -709,6 +714,7 @@ jobs:
         DEPLOYMENT: smoke
         WORKSPACE: release-((release_minor))-containerd
         TF_VAR_USE_CONTAINERD: "true"
```
aoldershaw, line 716:
```suggestion
        TF_VAR_ENABLE_CONTAINERD: "true"
```

tasks/terraform-smoke.yml
```diff
@@ -13,6 +13,7 @@ params:
   WORKSPACE: default
   USE_HTTPS: 'true'
   TF_VAR_USE_CONTAINERD: "false"
```
aoldershaw, line 15: Not that it really matters but

```suggestion
  TF_VAR_ENABLE_CONTAINERD: "false"
```

aoldershaw: Or just remove it? The default is already set in terraform

tasks/scripts/terraform-smoke
```diff
@@ -18,6 +39,13 @@ echo "$SSH_KEY" > keys/id_rsa
 chmod 0600 keys/id_rsa
 ssh-keygen -y -f keys/id_rsa > keys/id_rsa.pub
 
+if [[ $CONCOURSE_VERSION != "dev" ]]; then
+  cmp $CONCOURSE_VERSION "6.5.0"
```
izabelacg, line 43: Actually, do you have to `set +e` before this? Does the command "fail" if the return code isn't 0?

izabelacg: It looks like it: tried running the following script:

```sh
#!/bin/bash

set -e

cmp() {
  # strip pre-release (-rc) or build (+123) from semver
  v1=$(echo $1 | sed -e 's/[-+].*$//g')
  v2=$(echo $2 | sed -e 's/[-+].*$//g')

  if [[ $v1 == $v2 ]]; then
    return 0
  fi

  # use the magic of sort to figure out which version is smaller
  smaller=$(echo -e "$v1\n$v2" | sort -t "." -k 1,1 -k 2,2 -k 3,3 -g | head -n1)
  if [[ $smaller == $v1 ]]; then
    return 1
  fi
  return 2
}

if [[ 1 != 2 ]]; then
  cmp '1.2.3' '4.5.6'
  case $? in
    0) echo "=";;
    1) echo "<";;
    2) echo ">";;
  esac
fi
```

and it exited with error code 1 without `echo`ing anything. Adding `set +e` before the `cmp` and `set -e` after the `esac` allows it to work properly

chenbh: That's a good point, I think I'm gonna have it be based on output instead of exit codes, don't like the idea that we won't know if `sort` failed somehow.

Issue #344, opened 2020-07-07
regressions on older Linux kernels


jamieklassen:
# Problem

When garden-runc-release v1.19.11 was released, it broke compatibility with a specific edge case -- running inside a user namespace on a Linux 4.4 system. This is a rare scenario for Cloud Foundry, where guardian is run in the initial user namespace -- however it is not unusual for the concourse worker (and consequently guardian as well) to run in a container (= user namespace), like on Docker or k8s.

We didn't notice the impact this breaking change had on our users and actually shipped multiple subsequent releases before thoroughly investigating concourse/concourse#5711, and we still don't have a compelling workaround for users who cannot upgrade from Linux 4.4 or run in the initial user namespace.

If we plan to restore compatibility with the 4.4-user-namespace use case, the only path forward seems to be https://github.com/concourse/concourse/issues/5407 since garden is not being actively maintained.

In any case, how can we prevent gaffes like this in the future?

# Solution

Some subsystem of concourse -- the worker would be most relevant -- could get continuously exercised against older versions of Linux. The first idea that came to mind was:
1. add node pools similar to https://github.com/concourse/hush-house/blob/519ee144b103ce0678bca19b3a30ecd44dda54fd/terraform/main.tf#L77-L88 but running Linux 3.19 and 4.4. Add statefulsets for those workers to https://github.com/concourse/hush-house/tree/master/deployments/with-creds, with appropriately chosen [tags](https://github.com/concourse/hush-house/blob/519ee144b103ce0678bca19b3a30ecd44dda54fd/deployments/with-creds/ci-pr/values.yaml#L43) like `linux-3.19` and `linux-4.4`, for example.
1. add jobs to our pipelines which run the containerd integration tests on these workers, i.e. a `containerd-linux-4.4` and a `containerd-linux-3.19`.

## Questions
* does GKE provide machine images for these versions of Linux? If so, are they capable of also running kubernetes nodes? If _not_, do we manage the configuration for these workers elsewhere? maybe raw `google_compute_instance`s in terraform, provisioned similar to https://github.com/concourse/ci/blob/master/deployments/smoke/smoke.tf? If we manage these workers outside of k8s How do we meaningfully share config (like worker keys) between two config management systems?
[...]

PR #365, opened 2020-08-11
Add job to run bin smoke test in VM based on legacy kernel (4.4)

xtremerui:
Implements https://github.com/concourse/ci/issues/344

We can verify if the bin-smoke-legacy job runs on a VM with older kernel, the behaviour mentioned in https://github.com/concourse/ci/issues/344 could be reproduced. And the test will hang forever. Ideally we'd prefer a way to either fail the test and/or output the garden-runc error. But at least for now if the job hangs one would know something is wrong and it'd better to check the logs of the smoke VM.

@jamieklassen @xtreme-sameer-vohra @xtreme-vikram-yadav


tasks/terraform-custom-image.yml
```diff
@@ -0,0 +1,16 @@
+---
```
aoldershaw, line 1: Should this task be removed? I'm guessing it was replaced by `tasks/packer-custom-image.yml`?

tasks/packer-custom-image.yml
```diff
@@ -0,0 +1,20 @@
...
+params:
+  GCP_PROJECT:
+  GCP_KEY:
+
+inputs:
+- name: ci
+- name: linux-kernel-44
+
+outputs:
+- name: outputs
```
aoldershaw, line 17: Maybe use a less generic output name e.g. `packer-image`? And/or use `output_mapping` in the `task`?

tasks/scripts/packer-custom-image
```diff
@@ -0,0 +1,23 @@
...
+
+cd ci/deployments/legacy-kernel
+
+echo "$GCP_KEY" > gcp.json
+export GOOGLE_APPLICATION_CREDENTIALS=gcp.json
+
+packer build \
+  -var region="us-west1" \
+  -var machine_type="n1-standard-1" \
+  -var zone="us-west1-b" \
```
aoldershaw, line 20: What are the implications of the region and machine type? Does it need to match exactly with the size/location of the VM it's used for?

If so, does the terraform need to be updated? https://github.com/concourse/ci/blob/3200d4f2ab4fca3a3501d77852606957d3b76f2f/deployments/legacy-kernel/main.tf#L3-L9

tasks/scripts/packer-custom-image
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -e -u
+
+image_name="u16k44"
+outputs=$PWD/image-name
```
aoldershaw, line 6:
```suggestion
```

tasks/scripts/packer-custom-image
xtremerui: Interesting. The image that created in `us-west1` has a location `us`, from what I can see in gcp console. It seems out terraform smoke deployment that configured to `us-central1` can find it without problem. So I think as long as they are all located within US the regions/zones don't need to match.

aoldershaw: Cool, thanks. Also, is there a reason for going with `f1-micro`? Seems really small (1 vCPU, 0.6 GB memory)

xtremerui: this `main.tf` file is only used for testing if the packer created image could be used by terraform to create a VM. So it is for the script in https://github.com/concourse/ci/blob/15bc8429c18d9e894dae2b2c884e105d75120998/deployments/legacy-kernel/README.md.

Issue #297, opened 2020-03-23
The check params jobs rely on global state which breaks our main + release pipelines


xtreme-sameer-vohra:
This state in the CI repo (eg.  ci/tasks/scripts/check-distribution-env/helm/ignored-in-concourse) is global in that its shared across all the releases. 

We ran into a scenario with `CONCOURSE_GARDEN_USE_CONTAINERD`  in `tasks/scripts/check-distribution-env/helm/ignored-in-concourse` where it is desired on master (v6), however, it is not desired on the previous releases (v5.8.x) and its preventing us from shipping a patch release.

The CI repo, since it is global, should not have any state in it and these state files could be kept closer to the source ( in the packaging or concourse repo ). The CI repo would only have the scripts that act on these state files as they are common across all the releases

PR #367, opened 2020-08-17
Remove check-distribution-env state for CI checking

xtreme-sameer-vohra:
# Existing Issue
https://github.com/concourse/ci/issues/297

# Why do we need this PR?

Our older release pipelines fail because either the Concourse binary, BOSH release or Helm Chart have modified configurable values in the current release. As the state for which variables is currently global in the CI repo, this makes it hard to maintain multiple releases simultaneously.

# Changes proposed in this pull request

Remove [these ignored-* files](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env/helm) & [these ignored-* files](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env/bosh) from the CI repo to the Concourse Chart & BOSH release repo where they can be modified to suit each release.

# Reviewer Checklist
> This section is intended for the core maintainers only, to track review progress. Please do not
> fill out this section.
- [ ] Code reviewed
- [ ] Topgun tests run
- [ ] Back-port if needed


chenbh:
Is the intent to move the source-of-truth or the logic into their respective repos? 

If we want to move the logic we should also move the `tasks/scripts/check-distribution-env/*/list-actual` scripts as well (since they're the ones who knows how to read the template/props file. 

If we want to move the source-of-truth, the paths in the other repos should probably not be under scripts e.g. `ci/check-env-vars/ignored-in-*`

jamieklassen:
+1 to moving the logic as well -- these become a bit like "property-based tests" for the packaging itself, and ideally the sort of check that runs in the pipeline that gates shipping fully-packaged product, rather than the binaries/images inside.

xtreme-sameer-vohra:
> Is the intent to move the source-of-truth or the logic into their respective repos?
> 
The intent was to move the source-of-truth as that differs per release, but the logic is static. 
This has been an issue for a while and it would prevent us from shipping a patch for `release/6.3.x`. 

> If we want to move the logic we should also move the `tasks/scripts/check-distribution-env/*/list-actual` scripts as well (since they're the ones who knows how to read the template/props file).
> 
> If we want to move the source-of-truth, the paths in the other repos should probably not be under scripts e.g. `ci/check-env-vars/ignored-in-*`
>
Makes sense. What do you think of storing the files under `packaging-state` in the packaging repos instead ?

------
@jamieklassen  Moving the logic as well would include
- having a way to provide the concourse binary as an input to the test
- would be platform dependent

Would the tradeoff be worth it?

chenbh:
> Makes sense. What do you think of storing the files under packaging-state in the packaging repos instead ?

👍   
Maybe also leave a README in there for when unsuspecting people find this seemingly random list of env vars

xtreme-sameer-vohra:
> > Makes sense. What do you think of storing the files under packaging-state in the packaging repos instead ?
> 
> 👍
> Maybe also leave a README in there for when unsuspecting people find this seemingly random list of env vars

Done.

PR #375, opened 2020-09-02
Update scripts to use helm3

xtreme-vikram-yadav:
- Remove obsolete helm2 commands
 - Update concourse/unit image to use helm3

 Related to: https://github.com/concourse/concourse-chart/issues/56

Notes for Reviewer:

This PR depends on https://github.com/concourse/concourse-chart/pull/146.

### DIFF
--- a/dockerfiles/unit/Dockerfile
+++ b/dockerfiles/unit/Dockerfile
@@ -61,6 +61,12 @@ RUN curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add 
 # install helm for k8s-related jobs
 RUN curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash
 
+# TODO: tech-debt for maintaining 6.3.x version of the chart which uses Helm v2
+# Remove when 6.3.x no longer needs to be supported
+RUN curl https://get.helm.sh/helm-v2.16.12-linux-amd64.tar.gz -o helm-v2.tar.gz && \
+        tar xzf helm-v2.tar.gz && \
+        mv linux-amd64/helm /usr/local/bin/helm2
+
 # install Vault for bosh-topgun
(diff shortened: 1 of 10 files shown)


taylorsilva (changes requested):
One thing we haven't thought about is that these task files are also used by older `release/x.x.x` pipelines. Specifically I'm concerned with `release/6.3.x` which is our LTS version. We need to make sure we an continue to run this pipeline. 

Right now I'm seeing two possible paths:
1. Create a bunch of if-else branching in our tasks files to determine if `helm` should be v2 or v3. This incurs tech-debt and I don't favour this strategy.
1. Back-port all of these changes across all repos to `release/6.3.x`. Basically, repeat concourse/concourse-chart#56 for the `release/6.3.x` branches

I think it may be worth doing the second option. v2 will stop releasing security patches in November and we need to continue supporting 6.3.x until 2021.

wdyt?

xtreme-vikram-yadav:
Second option sounds good

YoussB:
@taylorsilva, @xtreme-vikram-yadav 

if I am a customer that is using the concourse 6.3.x chart with helm 2 and then upgrade the chart for a patch version of 6.3, wouldn't this affect my whole deployment as back-porting will make the chart only compatible with helm 3 🤔 ?

wdyt?

taylorsilva:
@YoussB agreed, it's an odd situation. It would be a very heavy patch for customers to upgrade too. Maybe we take the tech debt hit in this instance then and leave comments to remove it later?

taylorsilva:
@YoussB @xtreme-vikram-yadav I added a commit to ensure that k8s-topgun on 6.3.x still passes. Let me know what you think. [The commit I added](https://github.com/concourse/ci/pull/375/commits/6b5374767f18d63cdbaf04a103d26fd0b904c6e5). Read the commit message for full details.

Issue #369, opened 2020-08-17
enable `across` step on CI, use it for testflight/bin-smoke permutations


jamieklassen:
blocked on concourse/concourse#5887, which I just approved.

Let's see if we can't collapse the outlined sections of our development pipeline using the magic of build matrices:
<img width="990" alt="Screen Shot 2020-08-17 at 5 48 08 PM" src="https://user-images.githubusercontent.com/22056320/90447777-00c60780-e0b2-11ea-8ecb-af9820a6a299.png">

jamieklassen:
The stale bot scared me, so I did some work on this this morning -- now we have a single testflight job which matrixes across the worker runtime and postgres version!

zoetian:
Dev Notes: 

@chenbh and I picked up this issue. However, we did not realize the work on branch `issue/369`, we created a temporary branch `issue/369-bin-smoke` for the bin-smoke only. We will also port over the changes for `testflight/prs` and `testflights/release-pipelines` tomorrow.

chenbh:
This was done

PR #380, opened 2020-10-16
use across step for testflight, bin-smoke, bosh-topgun

chenbh:
#369 

There's already code on master by @jamieklassen to enable the across step for testflight on the concourse pipeline. This PR combines that with the work @aoldershaw did on enabling the steps for the pr and release pipelines.


chenbh:
I think this could be used for `bosh-smoke` and `bosh-smoke-containerd`, but there are some things I want to try out first: https://github.com/concourse/ci/blob/bfa091e4964ad676ba2e0b08a0eae31340cd300b/pipelines/concourse.yml#L801-L903

chenbh:
Turns out the bosh deployment resource (`put: smoke-deployment-containerd`)  requires the deployment name to be part of the source, working around this should be possible but is probably not worth the effort

pipelines/concourse.yml
```diff
@@ -289,25 +284,31 @@ jobs:
...
     image: unit-image
     privileged: true
     timeout: 1h
-    params:
-      CONTAINERD: ((.:enable-containerd))
-      POSTGRES_IMAGE: ((.:postgres-image))
+    params: 
+      RUNTIME: ((.:runtime))
+    input_mapping: 
+      postgres-image: postgres-image-((.:postgres))
```
aoldershaw, line 311: mentioned in standup, but just as a reminder, `input_mapping` doesn't support var interpolation, so you'll have to pass it in as a param and mount both inputs

PR #386, opened 2020-12-02
Add support to release arm64

electrocucaracha:
Given the latest popularity of arm64 is important to provide linux
binaries which are compiled against this architecture. These changes
add tasks to generate arm64 binaries for linux.


muntac:
@electrocucaracha Thanks for the PR! What devices are you planning to run the `linux/arm64` binaries of Concourse on?

electrocucaracha:
@muntac my end goal is to deploy it on top of k8s running on raspberry pi cluster

aoldershaw:
@martin-g also expressed some interest in working on arm64 support, and I gave a few notes on how to achieve that in the pipelines: https://github.com/concourse/concourse/discussions/6439#discussioncomment-299346

There are a fair number of things to consider, and I'm not totally sure about the best approach for everything, but hopefully this helps a bit if you're still interested in working on it

tibeer:
Adding myself here as I'm also interested in running concourse on top of Ampera ARM CPUs...

taylorsilva:
This PR isn't moving forward with the conflicts. There's also other stuff to think about, like packaging the resource types in arm images too, which isn't captured by this PR.

PR #387, opened 2020-12-14
check if docker mounts have changed

muntac:
This job checks whether the mount configuration in a regular and privileged Docker container have changed from
 our expectations.

 The mount configurations Concourse uses for containerd roughly follow Docker with some necessary exceptions. As a
 result we don't check against the containerd mounts directly.

 If this job fails:
 1. Find out why the changes were made in Docker https://github.com/moby/moby/
 Some files to check:
 https://github.com/moby/moby/blob/master/daemon/oci_linux.go
 https://github.com/moby/moby/blob/master/oci/defaults.go
 2. Determine if changes need to made to the mount configurations Concourse uses with containerd
 https://github.com/concourse/concourse/blob/master/worker/runtime/spec/mounts.go
 3. Finally update this file with the latest expected Docker mounts

Signed-off-by: Muntasir Chowdhury <muntasir.mzc@gmail.com>

### DIFF
--- a/tasks/docker-mounts.yml
+++ b/tasks/docker-mounts.yml
@@ -1,0 +1,17 @@
+---
+platform: linux
+
+image_resource:
+  type: registry-image
+  source: {repository: concourse/unit}
+
+params:
+  KUBE_CONFIG:
+  RELEASE_NAME:
+
+inputs:
+- name: ci
+
+run:
+  path: ci/tasks/scripts/start-docker
+  args: [ci/tasks/scripts/docker-mounts]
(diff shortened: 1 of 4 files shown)


tasks/docker-mounts.yml
```diff
@@ -0,0 +1,17 @@
+---
+platform: linux
+
+image_resource:
+  type: registry-image
+  source: {repository: concourse/unit}
+
+params:
+  KUBE_CONFIG:
+  RELEASE_NAME:
```
taylorsilva, line 10: Are these params used anywhere? Noticed they're not used in the new script and can't see how k8s-related envs could be useful here. Guessing this is leftover from a copy-paste?

taylorsilva (changes requested):

tasks/docker-mounts.yml
muntac: Yup, good catch!

taylorsilva (approved):

PR #394, opened 2021-03-19
Team tasks pipeline

taylorsilva:
This pipeline has two jobs:
- Rotate who organizes team lunch every month
- Rotate who runs standup everyday so @vito and @scottietremendous aren't always doing it!
- Rotate who runs retro every Friday

Not sure if there's a better way to do the rotating list thing. I made the resource a few weeks ago for the tutorial I'm slowly writing for the docs.

### DIFF
--- a/pipelines/team-tasks.yml
+++ b/pipelines/team-tasks.yml
@@ -59,6 +59,14 @@ resources:
     location: "America/Toronto"
     days: [Monday, Tuesday, Wednesday, Thursday, Friday]
 
+- name: friday-morning
+  type: time
+  source:
+    start: 11:00 AM
+    stop: 11:30 AM
+    location: "America/Toronto"
+    days: [Friday]
+
 - name: lunch
   type: static-list-resource
   source:
@@ -69,6 +77,12 @@ resources:
   type: static-list-resource
   source:
     unique: standup
+    list: *members
+
+- name: retro
+  type: static-list-resource
+  source:
+    unique: retro
     list: *members
 
 jobs:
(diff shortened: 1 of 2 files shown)


pipelines/team-tasks.yml
```diff
@@ -0,0 +1,108 @@
+members: &members
+# Aidan
+- "<@W013S2H8ABU>"
```
clarafu, line 3: Just curious what this hash is?

pipelines/team-tasks.yml
```diff
@@ -0,0 +1,108 @@
...
+- name: cron-resource
+  type: registry-image
+  source:
+    repository: cftoolsmiths/cron-resource
+
+resources:
+- name: slack-message
+  type: slack-notification
+  source:
+    url: ((slack_hook))
```
aoldershaw, line 45: nit: could we use `((slack_hooks.concourse-private))` instead? it's totally equivalent, but I've been trying to use `slack_hooks` where possible since it's explicit about what channel we send to

pipelines/team-tasks.yml
aoldershaw: looks like our slack member IDs

taylorsilva: yup, slack ID's!

pipelines/team-tasks.yml
taylorsilva: Will do! I blindly used what I saw in `concourse.yml`, so thanks for sharing!

pipelines/team-tasks.yml
```diff
@@ -0,0 +1,143 @@
...
+  - get: lunch
+  - put: slack-message
+    params:
+      text_file: lunch/item
+      text: |
+        $TEXT_FILE_CONTENT will organize this month's team lunch!
+        1. Set a date/time and create a calendar event
+        2. Think of something the team can do/watch (e.g. codenames)
+  - put: next-months-organizer
+    resourc: lunch
```
clarafu, line 102: Just a small spelling mistake!

```suggestion
    resource: lunch
```

clarafu (approved):
LGTM!!

PR #395, opened 2021-03-31
Rebuild and publish concourse images daily

taylorsilva:
This PR tries to bring us closer to image security best practices by rebuilding the concourse image for all active release branches ~and all resource images~ daily. Following [bitnami's footsteps here](https://docs.bitnami.com/tutorials/bitnami-best-practices-hardening-containers/#daily-builds-and-release-process), this is what they currently do.

~The resource pipeline has two jobs added; `shipit` for fanning in and `update-images` to build and republish both images on based on the last published version.~

The release pipeline only adds one job, `rebuild-images`. It builds images the same way the `build-rc-image` job does and publishes using the same steps the `publish-image` job does. The job has some additions:
- Along with the regular tags (e.g. `7.2.0`,`7.2`, etc) the images are also pushed to a tag that includes the date the image was pushed. The date tag is in the format of `7.2.0-{alpine,ubuntu}-YYYYMMDD`. This will allow users to opt-out of using the rolling tags (regular semver tags) if they want to, while clearly showing how old the image they are using is. Bitnami also does this with their images.
- Before publishing, the watsjs smoke test is run against both images
- As a sanity check:
  - verify that both images contain the expected binary version by comparing the output of `concourse --version` to the version tag that the images will be pushed to
  - scan both images with trivy to ensure all packages were updated with all available fixes

### DIFF
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -85,7 +85,7 @@ groups:
   - bump-cbd-versions
   - patch
   - discover-component-version
-  - update-images
+  - rebuild-images
 
 - name: all
   jobs:
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -85,6 +85,7 @@ groups:
   - bump-cbd-versions
   - patch
   - discover-component-version
+  - update-images
 
 - name: all
   jobs:
@@ -1342,6 +1343,8 @@ jobs:
     - get: version
       passed: [shipit]
     - get: linux-rc
+      passed: [shipit]
+    - get: linux-rc-ubuntu
       passed: [shipit]
     - get: concourse-docker
       trigger: true
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -1365,17 +1365,17 @@ jobs:
       output_mapping: {image: image-ubuntu}
       privileged: true
   - in_parallel:
-    - task: watsjs-ubuntu
+    - task: watsjs-smoke-ubuntu
       image: unit-image
       privileged: true
       timeout: 1h
-      file: ci/tasks/docker-compose-watsjs.yml
+      file: ci/tasks/docker-compose-watsjs-smoke.yml
       input_mapping: {dev-image: image-ubuntu}
-    - task: watsjs-alpine
+    - task: watsjs-smoke-alpine
       image: unit-image
       privileged: true
       timeout: 1h
-      file: ci/tasks/docker-compose-watsjs.yml
+      file: ci/tasks/docker-compose-watsjs-smoke.yml
       input_mapping: {dev-image: image-alpine}
   - load_var: image_version
     file: version/version
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -1340,6 +1340,9 @@ jobs:
   - in_parallel:
     - get: periodic-check
       trigger: true
+    - get: unit-image
+    - get: concourse
+      passed: [shipit]
     - get: version
       passed: [shipit]
     - get: linux-rc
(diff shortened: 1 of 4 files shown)


clarafu:
I just want to confirm with y'all that these changes make it so that we will constantly update the same concourse image with the most updated versions within the builder image. The only thing Im wondering is if we will ever want to keep a history of previous versions of the concourse images but built with older versions of the builder image?

taylorsilva:
> constantly update the same concourse image with the most updated versions within the builder image

Correct. I'm not sure if the way I've done it is the best way. I wondered if maybe having a one-line docker file that does
```dockerfile
RUN apt update && apt upgrade
```
on top of the published image would also be appropriate? Might give us more confidence that we're still shipping the same bits and aren't changing anything else.

pipelines/release.yml
```diff
@@ -1332,6 +1333,68 @@ jobs:
...
+      output_mapping: {image: image-alpine}
+      privileged: true
+    - task: build-ubuntu
+      file: concourse-docker/ci/build-image.yml
+      image: builder
+      input_mapping: {linux-rc: linux-rc-ubuntu}
+      output_mapping: {image: image-ubuntu}
+      privileged: true
+  - in_parallel:
+    - task: watsjs-ubuntu
```
vito, line 1368: rather than a full `watsjs` suite, would the smaller smoke tests suite make sense? (same tests we run in bin-smoke)

vito:
the PR descriptions mentions updating the resource pipeline, but I don't see that in the changes - was that pulled out?

vito:
@taylorsilva pulling on https://github.com/concourse/ci/pull/395#issuecomment-818044167 a little - I don't really have an intuition for what would be best here, but one other question that may help get us to an answer is whether we want to rebuild Concourse itself with new versions of Go that are released to address CVEs. that kind of seems valuable to me, but it would definitely be a more "intimate" (lol) daily-patch process as opposed to layering an 'apt-get update' on at the end.

pipelines/release.yml
taylorsilva: we did think about that! We chose watsjs because it specifically uses the image passed in. The smoke jobs simply install the binary on a vm and manage it with systemd, not using the image directly at all.

taylorsilva:
@vito updated the PR description. Gotta think about resources a bit more, that diff is on this branch: https://github.com/concourse/ci/tree/update-resource-images.

That is an interesting thought about rebuilding the binary with the latest go. I think in that case we would want to do more testing before re-releasing.

pipelines/release.yml
taylorsilva: so I dug into this a bit more and learned that the bin-smoke job you referred to runs a subset of the watjs smoke tests. Not sure if this is what you meant vs the entire suite. I added a task task that runs only the smoke tests in the watsjs suite.

pipelines/release.yml
```diff
@@ -85,6 +85,7 @@ groups:
   - bump-cbd-versions
   - patch
   - discover-component-version
+  - update-images
```
vito, line 88: guessing this is a typo/old name:

```suggestion
  - rebuild-images
```

vito (commented):
lgtm in principle but noticed a typo i think!

taylorsilva:
I found another typo as well. Adding image scanning to the list of sanity checks to run before publishing. Anything else that you feel hesitant about Alex or Clara?

pipelines/release.yml
vito: yep that's what i was going for - thanks! just figured it'll save some time since it's really just a sanity check

vito (approved):
lgtm!

PR #396, opened 2021-04-09
Remove hardcoding of dev-image repository

BooleanCat:
With this hard coding in place, `testflight` and `integration` tests will always use the `dev-image` produced by concourse' own CI. This ensures that dev-images produced by others can be used by this task.

### DIFF
--- a/tasks/scripts/integration
+++ b/tasks/scripts/integration
@@ -12,34 +12,33 @@ start_docker
 
 image_ref () {
   dir=$1
-  repo=$2
 
   # note: tag must take precedence, since the digest returned by the
   # registry-image resource is not the same as the image id in the docker CLI.
   # the oci-build-task, however, gives a digest that can be used
   if [ -f "${dir}/tag" ]; then
-    echo "${repo}:$(cat ${dir}/tag)"
+    echo "$(cat ${dir}/repository):$(cat ${dir}/tag)"
   else
     cat ${dir}/digest
   fi
 }
 
 docker load -i dev-image/image.tar
-export TEST_CONCOURSE_DEV_IMAGE="$(image_ref dev-image concourse/dev)"
+export TEST_CONCOURSE_DEV_IMAGE="$(image_ref dev-image)"
 
 if [ -d concourse-image ]; then
   docker load -i concourse-image/image.tar
-  export TEST_CONCOURSE_LATEST_IMAGE="$(image_ref concourse-image concourse/concourse)"
+  export TEST_CONCOURSE_LATEST_IMAGE="$(image_ref concourse-image)"
 fi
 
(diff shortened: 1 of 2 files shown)


tasks/scripts/integration
```diff
@@ -25,7 +25,7 @@ image_ref () {
 }
 
 docker load -i dev-image/image.tar
-export TEST_CONCOURSE_DEV_IMAGE="$(image_ref dev-image concourse/dev)"
+export TEST_CONCOURSE_DEV_IMAGE="$(image_ref dev-image "$(cat dev-image/repository)")"
```
vito, line 28: Makes sense, though this assumes the `repository` file exists - it won't if the input was generated by the `oci-build-task` (i.e. in a PR build).

I think we can push this logic into `image_ref`:

```suggestion
export TEST_CONCOURSE_DEV_IMAGE="$(image_ref dev-image)"
```

and then change L21 in `image_ref` to this:

```bash
    echo "$(cat ${dir}/repository):$(cat ${dir}/tag)"
```

L23 in `image_ref` seems a little sketchy, but that can be out of scope of this PR. It looks like it just sets the value to a naked digest, which doesn't work with `docker-compose` on my machine. Which begs the question - how is the PR build passing? 🤔 Maybe we can change it so that it gives the digest a random local name and then returns `${local_name}:$(cat ${dir}/digest)`. (Again, just thinking out loud, no need to address this here.)

vito (changes requested):
thanks! left a suggestion inline. the rest LGTM though.

tasks/scripts/integration
BooleanCat: I've pushed the repository lookup into `image_ref`, this also allowed us to remove the second argument for all calls to `image_ref`.

tasks/scripts/integration
```diff
@@ -12,34 +12,33 @@ start_docker
...
 image_ref () {
   dir=$1
-  repo=$2
 
   # note: tag must take precedence, since the digest returned by the
   # registry-image resource is not the same as the image id in the docker CLI.
   # the oci-build-task, however, gives a digest that can be used
   if [ -f "${dir}/tag" ]; then
-    echo "${repo}:$(cat ${dir}/tag)"
+    echo "$(cat dev-image/repository):$(cat ${dir}/tag)"
```
vito, line 20:
```suggestion
    echo "$(cat ${dir}/repository):$(cat ${dir}/tag)"
```

vito (changes requested):

tasks/scripts/integration
BooleanCat: 😬  Fixed!

vito (approved):
lgtm - thanks!

PR #397, opened 2021-04-09
add image vulnerability scanning to release pipelines

chenbh:
- using https://github.com/aquasecurity/trivy since it seems to be the most popular scanner that supports CLI-only
- we can use the official `aquasec/trivy` image or the `bitnami/trivy` fork

### DIFF
--- a/pipelines/resources/template.yml
+++ b/pipelines/resources/template.yml
@@ -95,6 +95,57 @@ plan:
   params: {image: built-resource-image/image.tar}
 #@ end
 
+#@ def scan_task(distro):
+task: #@ "scan-" + distro
+image: trivy
+config:
+  platform: linux
+  inputs:
+  - name: #@ "resource-image-dev-" + distro
+    path: image
+  - name: trivy-db
+  run:
+    path: sh
+    args:
+    - -c
+    - |
+      mkdir db
+      tar -xzf trivy-db/trivy-offline.db.tgz -C ./db
+
+      trivy \
+        --cache-dir $(pwd) \
+        image \
+        --severity "HIGH,CRITICAL" \
+        --ignore-unfixed \
+        --exit-code 1 \
+        --input image/image.tar
+#@ end
+
+#@ def scan_image():
+name: scan-image
+on_failure: &failed-ci
+  put: notify
+  params:
+    mode: normal
+    alert_type: failed
+plan:
+- in_parallel:
+  - get: resource-repo
+  - get: resource-image-dev-alpine
+    passed: [build-alpine]
+    params: {format: oci}
+  - get: resource-image-dev-ubuntu
+    passed: [build-ubuntu]
+    params: {format: oci}
+  - get: trivy
+    trigger: true
+  - get: trivy-db
+    trigger: true
+    params: {globs: [trivy-offline.db.tgz]}
+- #@ scan_task("alpine")
+- #@ scan_task("ubuntu")
+#@ end
+
 #@ def extra_steps(distro):
   #@ if resource_name == "docker-image":
 - task: smoke-test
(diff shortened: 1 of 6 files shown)


aoldershaw (commented):
Trivy seems really cool! Rather than just scanning the image when we generate an RC, I wonder if we should continuously scan the latest RC against the Trivy DB separately - i.e. have a separate job for vulnerability scanning which triggers on changes to the RC images and on new releases in https://github.com/aquasecurity/trivy-db/releases

I guess it doesn't have to be a separate job, but it might be nice to split it up so we aren't triggering the rest of the pipeline when we get a new vulnerability DB and so the notifications we get are more telling.

Not totally necessary - we're effectively scanning everything daily anyway, since we build a new RC daily. Just an idea!

taylorsilva:
I thought about adding trivy as part of https://github.com/concourse/ci/pull/395 but decided not to for the reason Aidan gave. If you release daily then there's no need to scan because you will have pulled in the latest fixes. Even if there is a CVE but there's no fix for it, the best thing you can always do is re-build. That's why I ended up not including trivy in my PR. I'm not sure if it's worth failing and blocking us on a CVE when no fix is available.

chenbh:
@aoldershaw I worded the title poorly, the `build-rc` task is meant to be more of a sanity check. I'm thinking of adding a separate job to scan the actual released image instead of the rc, as well as adding similar scanning to the resource pipelines. 

I actually don't like the idea of "pulling in the latest version from `apt-get`", it breaks the concept of reproducible builds and makes it hard to audit an image. I'm thinking that in the future, we should have some sort of `dpkg.lock` file that can definitively tell us what dependencies to install. That way, every version bump is an explicit action from us (with the help of dependabot). 

The vulnerability scanning task/job can be also be used as an early warning/short-circuit; it would be nice for us to preemptively warn the community or customers of a high-level CVE instead of silently bumping the image behind the scenes.

@taylorsilva that's what the `--ignore-unfixed` flag is for

chenbh:
Also, both the `concourse-rc` and `concourse-rc-ubuntu` are built from [the same dockerfile](https://github.com/concourse/concourse-docker/blob/master/Dockerfile), it might make sense to only scan one of them (the actual difference is in the os of the base resource types they come with).

aoldershaw:
@chenbh yeah, I like the idea of continuously scanning the released image, and scanning the RC prior to releasing is a good sanity check.

> Also, both the concourse-rc and concourse-rc-ubuntu are built from the same dockerfile, it might make sense to only scan one of them (the actual difference is in the os of the base resource types they come with).

Do we know if trivy will recurse into the tarballs for the base resource types? My guess is no, but I'm not sure. In any event, we should probably scan all of the resource types separately for each of `concourse-rc` and `concourse-rc-ubuntu` as well. It probably makes sense to do this in the release pipeline (since we pin the versions down after releasing), but maybe also in the resource type pipelines (or `concourse` pipeline) as well so that we're constantly tracking the latest versions?

chenbh:
@aoldershaw I've split the scanning into a separate job so we can trigger off of `aquasecurity/trivy-db`, as well adding resource-types scanning to both the release and resource pipelines.

pipelines/resources/template.yml
```diff
@@ -177,24 +228,22 @@ name: #@ "publish-" + bump
 plan:
 - in_parallel:
   - get: resource-repo
-    passed:
-    - build-alpine
-    - build-ubuntu
```
chenbh, line 182: this technically isn't needed, but it makes the graph look a lot nicer

aoldershaw (approved):
Very cool! Was thinking it might make sense to add this to the main (concourse) pipeline as well so that we can proactively find/fix vulnerabilities, but in hindsight it probably doesn't make sense - we're always rolling forward, and this is likely mainly going to catch issues with older base images

PR #399, opened 2021-04-26
Pin release pipeline's resource types to minor version that was released

aoldershaw:
The approach taken was:
* Encode the released version as a semver constraint in the `reconfigure` pipeline
  * I considered checking the versions into the repo in its own file, which would make the pinning portion less wonky, but wanted everything to be self-contained (so that adding a new release pipeline is just as easy)
  * note: `~1.2.3` means versions `>= 1.2.3` and `<1.3.0`, i.e. patch versions at least as new as `1.2.3`
* Extract the resource type versions for generating the release notes and for pinning
  * We previously extracted the versions, but encoded them directly into the release notes - now, there's a separate `extract-resource-type-versions` task, and the release note generation is all located within `build-release-notes`
* After `shipit`, there's a new `pin-resource-type-versions` updates `pipelines/reconfigure.yml` with the bumped versions
  * This is done in a python scripts that's a bit hacky - I didn't want to decode the YAML directly, since when we re-marshal it, we'd lose stuff like comments, so instead, we walk the lines looking for `set_pipeline` steps, and when we find the correct one, replace it

Note: blocked on concourse/github-release-resource#105


pipelines/reconfigure.yml
```diff
@@ -333,6 +333,21 @@ jobs:
...
+          bosh-io-release: ~1.0.3
+          bosh-io-stemcell: ~1.0.3
+          cf: ~1.1.0
+          docker-image: ~1.5.1
+          git: ~1.12.1
+          github-release: ~1.5.1
+          hg: ~1.2.0
+          mock: ~0.11.0
+          pool: ~1.1.4
+          registry-image: ~0.14.0
```
chenbh, line 346: I'm not sure this will be too useful without setting up the ability to release patch for previous resource-type minors. With the 7.x series we can reasonably say just bump to the latest version, but for 6.7.x there's no way for us to release a patch of the registry-image-resource even if we know it has vulnerabilities

aoldershaw: Hm yeah that's true - if we ever get to a point where we can support multiple versions of the resource types simultaneously, this'll be more useful. It still does have *some* value now, in that we won't need to manually pin all the resources after releasing

tasks/scripts/build-release-notes
```diff
@@ -12,8 +14,20 @@ chmod +x release-me/releaseme
...
   --ignore-authors=dependabot \
-  > built-notes/notes.md
+  > $FILE
+
+if [ -f resource-type-versions/versions.yml ]; then
+  cat >> $FILE <<EOF
+
+## 📦 Bundled resource types
+
+<details>
```
chenbh, line 24: This will need a closing `</details>` at the end

tasks/scripts/build-release-notes
```diff
@@ -12,8 +14,20 @@ chmod +x release-me/releaseme
...
 
-if [ -f resource-versions/versions ]; then
-	cat resource-versions/versions >> built-notes/notes.md
+EOF
+  cat resource-type-versions/versions.yml | while read line; do
+    name=$(printf $line | cut -d':' -f1 | xargs)
+    version=$(printf $line | cut -d':' -f2 | xargs)
+    ghlink="https://github.com/concourse/$name-resource/releases/tag/$version"
+    echo "- $name: [$version]($ghlink)" >> $FILE
+  done
```
chenbh, line 32: Harness the arcane powers of `awk`:

```
awk 'BEGIN {FS = ": "}; { printf "- %s: [%s](https://github.com/concourse/%s/releases/tag/%s)\n", $1, $2, $1, $2 }' resource-type-versions/versions.yml
# or slightly more readable:
awk 'BEGIN {FS = ": "}; { url = sprintf("https://github.com/concourse/%s/releases/tag/%s", $1, $2); printf "- %s: [%s](%s)\n", $1, $2, url }' resource-type-versions/versions.yml
```

aoldershaw: nice, haven't really bothered to learn `awk` effectively yet but seems pretty useful - cleaned up in 1749c3d1e0dd103f198ecf9cfae847ac6100c4f3

chenbh (approved):

pipelines/release.yml
```diff
@@ -1295,6 +1296,24 @@ jobs:
...
+    - get: ci
+  - task: extract-resource-type-versions
+    file: ci/tasks/extract-resource-type-versions.yml
+  - task: pin-resource-type-versions
+    file: ci/tasks/pin-resource-type-versions.yml
+    params:
+      RELEASE_MINOR: ((release_minor))
+    output_mapping: {ci: bumped-ci}
+  - put: ci
+    params: {repository: bumped-ci}
```
chenbh, line 1315: @aoldershaw looks like the private key in vault doesn't have push permissions: https://ci.concourse-ci.org/teams/main/pipelines/release/jobs/pin-resource-type-versions/builds/3?vars.version=%227.2.x%22

aoldershaw: Thanks, fixed now

PR #400, opened 2021-04-27
Capture stemcell version in release pipelines

taylorsilva:
When releasing an enterprise version of Concourse we need to know what
stemcell version we tested against. This commit captures that info and
saves it in gcs for use by the enterprise pipeline on runway.


clarafu (approved):
Cool! LGTM :D

PR #403, opened 2021-07-09
Set an instanced pipeline per PR

aoldershaw:
This makes use of the new dynamic `across` behaviour in concourse/concourse#7252 (so is therefore blocked on it). It makes the following changes:

* `reconfigure-pipelines` is responsible for tracking changes to the list of PRs to the concourse repo (to the `master` branch only) and sets an instanced pipeline for each one
* Each `release` pipeline also tracks changes to the list of PRs to their corresponding `release-M.m.x` branch, and does the same
* The PR pipelines themselves are different. Since we no longer have to worry about `version: every` + `passed` constraints, we can have a proper "pipeline" that builds the image once and passes it along to downstream jobs:
Screen Shot 2021-07-09 at 9 27 40 AM

There were a few other changes related to the across step I added:
* Simplify the setting of the resource pipelines
* Make use of the new across substep interpolation in concourse/concourse#7252 to interpolate the `input_mapping` - this means we no longer need to pass in both `postgres-image` and `postgres-image-legacy` to each `testflight` container, and can just pass in the one we're actually using.


pipelines/pr.yml
```diff
@@ -86,39 +126,62 @@ resources:
...
+      platform: linux
+      params:
+        CONTEXT: built-concourse
+        IMAGE_ARG_base_image: dev-image/image.tar
+        TARGET: with-fly
+      inputs: [{name: built-concourse}, {name: dev-image}]
+      outputs: [{name: image}]
+      # TODO: now that this is per PR pipeline, caches might end up taking a
+      # ton of disk...figure out if this is a problem.
+      caches: [{path: cache}]
```
aoldershaw, line 162: This is what I was talking about in retro. I think without some sort of GC mechanism for task caches we'll run into issues, since the size of the `pr` instance group never actually shrinks (old PR pipelines will get archived, but it seems as though task caches remain for archived pipelines)

aoldershaw: No longer blocked (see https://github.com/concourse/concourse/pull/7272)

PR #405, opened 2021-09-21
Build resource-types image from docker hub

taylorsilva:
instead of building from github release artifacts, which are hard to update, build from the images that we constantly push to docker hub. This allows us to not make silly patch releases for simply updating the OS packages in the resource images.

Closes concourse/concourse#7053

### DIFF
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -177,56 +177,115 @@ jobs:
   serial: true
   plan:
   - in_parallel:
-    - get: bosh-io-release-resource
-      trigger: true
-    - get: bosh-io-stemcell-resource
-      trigger: true
-    - get: cf-resource
-      trigger: true
-    - get: docker-image-resource
-      trigger: true
-    - get: git-resource
-      trigger: true
-    - get: github-release-resource
-      trigger: true
-    - get: hg-resource
-      trigger: true
-    - get: pool-resource
-      trigger: true
-    - get: registry-image-resource
-      trigger: true
-    - get: s3-resource
-      trigger: true
-    - get: semver-resource
-      trigger: true
-    - get: time-resource
-      trigger: true
-    - get: tracker-resource
-      trigger: true
-    - get: mock-resource
-      trigger: true
     - get: builder
       trigger: true
     - get: periodic-check
       trigger: true
     - get: ci
   - in_parallel:
-    - task: build-alpine
-      image: builder
-      privileged: true
-      params:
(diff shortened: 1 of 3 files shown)


tasks/build-resource-types-image.yml
```diff
@@ -29,4 +28,20 @@ inputs:
 outputs:
 - name: image
 
-run: {path: build}
+run:
+  path: bash
```
taylorsilva, line 32: confirmed the `concourse/builder` does have bash and tar in the image

pipelines/release.yml
```diff
@@ -1965,131 +2024,257 @@ resources:
...
+  type: registry-image
   icon: *release-icon
   source:
-    owner: concourse
-    repository: mock-resource
-    access_token: ((concourse_github_dummy.access_token))
-    semver_constraint: ((resource_type_versions.mock))
+    repository: concourse/mock-resource
+    tag: ((resource_type_versions.mock))-alpine
+    username: ((docker.username))
```
clarafu, line 2033: Do we need to give it a username and password if we are just pulling from docker hub?

clarafu (approved):
Thanks for doing this haha will be super useful!

pipelines/release.yml
taylorsilva: Only reason I added it was to avoid any rate limit issues since this means we're pulling a lot more images. Based on https://www.docker.com/increase-rate-limits it goes from 100 to 200 for auth'd requests.

clarafu:
We are seeing errors in our ci for the `create resource config: base resource type not found: mock` https://ci.concourse-ci.org/teams/main/pipelines/release/jobs/testflight/builds/40?vars.version=%227.5.x%22 and also `cat: can't open 'resource-types/github-release/resource_metadata.json': No such file or directory
` https://ci.concourse-ci.org/teams/main/pipelines/release/jobs/build-rc/builds/45?vars.version=%227.5.x%22

This is caused by the `metadata.json` containing what looks like runtime information 
image
versus the `resource_metadata.json` that contains the resource type information 
image

taylorsilva:
Fixed by the following commits
- https://github.com/concourse/ci/commit/57403133da6b5cb71984e8900be713c2477626c6
- https://github.com/concourse/ci/commit/4447312efe76c8aa71d0ff731e21be027ee89cc6
- https://github.com/concourse/ci/commit/47742d97d4047e29d7fe734bc4fc336a2a2a04e3

PR #407, opened 2022-01-27
add linux-ubuntu to the build binary file name

deepakmohanakrishnan07:
Currently both linux and ubuntu file names are identical which can potentially
overwrite the file by `put:linux-rc` and `put:linux-rc-ubuntu` steps. To avoid it,
adding ubuntu to the binary file name and also match the file name pattern
defined in `linux-rc-ubuntu` resource.

### DIFF
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -656,7 +656,7 @@ jobs:
         output_mapping: {concourse-tarball: concourse-linux-alpine}
       - task: concourse-linux-ubuntu
         image: golang-builder-image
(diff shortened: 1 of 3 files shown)


xtremerui:
Hi, `PLATFORM` is referring to the output of `go env GOOS` i.e. the OS that golang supports for cross compile. `linux-ubuntu` is not a valid candidate. In our case we'd want to keep the param to align with `linux` `darwin` and `windows`.

taylorsilva:
Wasn't seeing a reason to keep this open

PR #420, opened 2025-02-04
add k8s-test job to test concourse helm chart installation

yharish991:
This PR adds a `k8s-test` to `helm-chart` pipeline to install and validate concourse helm chart before publishing it for confidence


tasks/scripts/k8s-test
```diff
@@ -0,0 +1,26 @@
...
+
+helm upgrade --install concourse-chart-test concourse-chart/ \
+  --wait \
+  --recreate-pods \
+  --namespace concourse-chart-test \
+  --create-namespace
+
+kubectl \
+  rollout status deployment \
+  concourse-chart-test-web --namespace concourse-chart-test
```
taylorsilva, line 26: This is a neat command! To be clear, it'll `exit 1` if the deployment fails?

tasks/scripts/k8s-test
taylorsilva, line 1: I'm not seeing any cleanup steps happen here. Do we just leave the cluster running indefinitely on that target GKE cluster?

taylorsilva (changes requested):

tasks/scripts/k8s-test
yharish991: yes, you are correct if the deployment fails, the above command returns a non-zero exit code to indicate the failure

tasks/scripts/k8s-test
yharish991: currently, it doesn't do any cleanup, i think it is okay to leave running as the next task run would upgrade the existing installation and that shouldn't be a problem for subsequent runs. However, i think it would be ideal to do the clean up

yharish991: Oh didn't notice that we had k8s-smoke job, let me see if we can re-use it, good catch :)

taylorsilva (approved):

taylorsilva:
Made some small changes, looks good.

PR #423, opened 2025-04-11
Wolfi base

taylorsilva:
Related to https://github.com/concourse/concourse/issues/9183


taylorsilva:
I think I'm going to hack away at the registry-image now and then merge this in. I expect things to be broken when I merge this since testing is hard. Top priority will be fixing whatever I broke here.

taylorsilva:
Will be fixing stuff in-flight now.

PR #424, opened 2025-06-01
fix: Update build-resource-types-image/run.sh

kcbimonte:
Switch build-resource-types-image/run.sh to use sh instead of bash due to OCI Build Task not having bash

### DIFF
--- a/tasks/build-resource-types-image/run.sh
+++ b/tasks/build-resource-types-image/run.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 set -euo pipefail
 


taylorsilva (commented):
Thanks for the PR! I'm not sure this will work though. I quickly ran this through shellcheck and it raised some concerns like:

```
In POSIX sh, set option pipefail is undefined.

In POSIX sh, [[ ]] is undefined.
```

There was a bunch of other stuff too, unrelated to the bash -> sh change though.

I'm doing a lot of work in this repo right now too related to arm64 release building. I'm working on the job that uses this script right now. I'm just circle back and resolve the TODO I left in the script here and update the oci-build task to use wolfi now.

PR #431, opened 2026-07-20
extend testing on windows

aliculPix4D:
extend testing on windows

- go-archive unit test must also run on windows platform

Sadly, I don't know how to test this properly from the Concourse CI side. 

Additionally, there is 1 test failing: https://github.com/concourse/concourse/blob/master/go-archive/tarfs/extract_test.go#L221 that requires a PR in Concourse repo (in preparation)

### DIFF
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -140,7 +140,6 @@ jobs:
             image: unit-image
             file: ci/tasks/unit/task.yml
           - task: unit-windows
-            attempts: 3
             file: ci/tasks/unit-windows/task.yml
 
   - name: unit-yarn
--- a/pipelines/concourse.yml
+++ b/pipelines/concourse.yml
@@ -134,10 +134,14 @@ jobs:
           - get: unit-image
             trigger: true
           - get: ci
-      - task: unit
-        timeout: 1h
-        image: unit-image
-        file: ci/tasks/unit/task.yml
+      - timeout: 1h
+        in_parallel:
+          - task: unit
+            image: unit-image
+            file: ci/tasks/unit/task.yml
+          - task: unit-windows
+            attempts: 3
+            file: ci/tasks/unit-windows/task.yml
 
   - name: unit-yarn
     public: true
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -139,7 +139,6 @@ jobs:
             image: unit-image
             file: ci/tasks/unit/task.yml
           - task: unit-windows
-            attempts: 3
             file: ci/tasks/unit-windows/task.yml
 
   - name: unit-yarn
--- a/pipelines/release.yml
+++ b/pipelines/release.yml
@@ -133,10 +133,14 @@ jobs:
           - get: periodic-check
             trigger: true
           - get: ci
-      - task: unit
-        timeout: 1h
-        image: unit-image
-        file: ci/tasks/unit/task.yml
+      - timeout: 1h
+        in_parallel:
+          - task: unit
+            image: unit-image
+            file: ci/tasks/unit/task.yml
+          - task: unit-windows
+            attempts: 3
+            file: ci/tasks/unit-windows/task.yml
 
   - name: unit-yarn
     public: true
--- a/tasks/unit-windows/run.ps1
+++ b/tasks/unit-windows/run.ps1
@@ -10,6 +10,6 @@ if ($LASTEXITCODE) { Throw "failed to determine Ginkgo version (exit code $LASTE
 go install "github.com/onsi/ginkgo/v2/ginkgo@$GinkgoVersion"
 if ($LASTEXITCODE) { Throw "Ginkgo installation failed (exit code $LASTEXITCODE)" }
 
-ginkgo -r -p ./go-archive
+ginkgo -r -p -flake-attempts=3 -race ./go-archive
 
 Exit $LastExitCode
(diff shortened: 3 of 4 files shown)


tasks/unit-windows/run.ps1
```diff
@@ -0,0 +1,15 @@
...
+
+go mod download
+
+$GinkgoVersion = go list -f '{{.Version}}' -m github.com/onsi/ginkgo/v2
+if ($LASTEXITCODE) { Throw "failed to determine Ginkgo version (exit code $LASTEXITCODE)" }
+
+go install "github.com/onsi/ginkgo/v2/ginkgo@$GinkgoVersion"
+if ($LASTEXITCODE) { Throw "Ginkgo installation failed (exit code $LASTEXITCODE)" }
+
+ginkgo -r -p ./go-archive
```
aliculPix4D: There are already: `unit-baggageclaim-windows` and `unit-fly-windows`. So I think the only one missing the tests on Windows is `go-archive`.

pipelines/concourse.yml
```diff
@@ -134,10 +134,14 @@ jobs:
...
-        timeout: 1h
-        image: unit-image
-        file: ci/tasks/unit/task.yml
+      - timeout: 1h
+        in_parallel:
+          - task: unit
+            image: unit-image
+            file: ci/tasks/unit/task.yml
+          - task: unit-windows
+            attempts: 3
```
marco-m-pix4d: For my understanding, why do we need the `attempts` ?

tasks/unit-windows/run.ps1
```diff
@@ -0,0 +1,15 @@
+. .\ci\tasks\scripts\go-build.ps1
+
+cd .\concourse
+
+go mod download
+
+$GinkgoVersion = go list -f '{{.Version}}' -m github.com/onsi/ginkgo/v2
+if ($LASTEXITCODE) { Throw "failed to determine Ginkgo version (exit code $LASTEXITCODE)" }
+
+go install "github.com/onsi/ginkgo/v2/ginkgo@$GinkgoVersion"
```
aliculPix4D, line 10: Maybe it's worthwhile to replace the .bat script with powershell in: https://github.com/concourse/ci/blob/master/tasks/unit-baggageclaim-windows/run.bat and do the same for ginkgo version to ensure we are installing the intended version.

pipelines/concourse.yml
aliculPix4D: We don't. Wrong copy paste... I will remove.

UPDATE: or given that we have this `attempts` almost everywhere else, let's see what @taylorsilva  says about it before removing.

aliculPix4D: I copy pasted from:

```
     - task: unit-baggageclaim-windows
       attempts: 3
       file: ci/tasks/unit-baggageclaim-windows/task.yml
```
and

```
      - task: fly-windows
        attempts: 3
        file: ci/tasks/fly-windows/task.yml
```

taylorsilva: These `attempts` were from us trying to work around flakey tests. The tests rarely flake now but still sometimes do.

I think we can remove it from here though.

tasks/unit-windows/run.ps1
taylorsilva: I'd also add these flags `-flake-attempts=3 -race`
```suggestion
ginkgo -r -p -flake-attempts=3 -race ./go-archive
```
Copied from: https://github.com/concourse/ci/blob/51425ef46e4904494a52b08de3e1fa60652c5055/tasks/unit/run.sh#L25

pipelines/release.yml
```diff
@@ -133,10 +133,14 @@ jobs:
...
-        timeout: 1h
-        image: unit-image
-        file: ci/tasks/unit/task.yml
+      - timeout: 1h
+        in_parallel:
+          - task: unit
+            image: unit-image
+            file: ci/tasks/unit/task.yml
+          - task: unit-windows
+            attempts: 3
```
taylorsilva:
```suggestion
```
I'd remove this since we can use ginkgo's `-flake-attempts` flag instead.

tasks/unit-windows/task.yml
taylorsilva, line 1: I'm `fly execute`'ing this task on our Windows worker to verify that this will work.

tasks/unit-windows/run.ps1
taylorsilva: I agree. I'll do this on `master` outside this PR.

tasks/unit-windows/task.yml
taylorsilva: It works! One of the test fails

```
  [FAILED] Expected
      <string>: \\absolute\\path\\file
  to equal
      <string>: /absolute/path/file
  In [It] at: C:/work/containers/00003tjkltd/tmp/build/b1dda4c3/concourse/go-archive/tarfs/extract_test.go:233
```

But that's technically unrelated to this PR. We'll have to fix this in `concourse/concourse`.

taylorsilva (approved):

tasks/unit-windows/run.ps1
aliculPix4D: @taylorsilva  Thanks for taking care of this PR. FYI I noticed the following:

Did my suggestion break the pipeline? https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/unit-baggageclaim/builds/375

> start process: backend error: Exit status: 500, message: {"Type":"","Message":"create process: %1 is not a valid Win32 application.","Handle":"","ProcessID":"","Binary":""}

or is it just a race?

This is the difference I am seeing.

In `tasks/unit-windows/task.yml` I did:
```yaml
run:
  path: powershell
  args: 
  - ci/tasks/unit-windows/run.ps1
```
while the updated `tasks/unit-baggageclaim-windows/task.yml` task in https://github.com/concourse/ci/commit/c342e80ad6687e8da32b0e3534965c988e876d81 commit has:
```yaml
run:
 path: ci/tasks/unit-baggageclaim-windows/run.ps1
```

aliculPix4D: PR: https://github.com/concourse/ci/pull/432

